PR #67, opened 2015-08-21
Add ffmpeg package and library names for ubuntu 15.04

c42f:
Add the ffmpeg versions of the libraries because the libav versions available
on ubuntu 15.04 have not yet wrapped in VideoIO.

The ffmpeg libraries have been put first in the search order so that they'll be
found in preference, even if the user also has the libav versions installed.
This avoids a VideoIO crash at runtime when loading libraries which haven't
been wrapped yet.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -5,12 +5,12 @@ using BinDeps
 group = library_group("libav")
 
 deps = [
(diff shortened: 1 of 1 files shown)


c42f:
This works around #66 for usage on ubuntu 15.04, but doesn't fix the underlying cause of the crash in #66.

c42f:
A note about the diff which is a bit confusing:  All I've done to the library alias lists is append `libav$foo-ffmpeg.so.` to each of the alias lists.

kmsquire:
Hi Chris, thanks for the PR. I'm on vacation right now, but will try to look this weekend.

PR #71, opened 2015-10-16
Update playvideo to work with v0.4 and latest ImageView

bnels:
Fixes problem where read! could not find an appropriate method.


kmsquire:
Sorry not to respond to this sooner (I've been mostly MIA on this package).

`playvideo` seems to work (for me) on the current master, without this change.  If you're interested in testing this, I'd love to hear if it doesn't work for you still.

Note that, if all goes as planned, the parts of this package which depend on `Images.jl` will be split out into a separate package at some point soon.  (Of course, that assumes I have time to work on this...)

Closing, but please post back here if this is still an issue, and I'll reopen.

PR #73, opened 2015-10-31
More 0.4 deprecation fixes

mbauman:
This is a fly-by PR with a simple global find and replace all for:

`\b(U?)int(\d*)\b` -> `$1Int$2` (There are a gazillion of these, I only did a few spot checks)
`\bString\b` -> `AbstractString` (These were all manually vetted)
`\bUnion\(([^)]*)\)` -> `@compat(Union{$1})` (These were all manually vetted. This doesn't work for nested parentheses, but there weren't any.)

And fix one macro definition whose parentheses were detached from the name.

Note that I did not use the wrapper scripts; this means that if Clang.jl isn't updated to use the new names the patches will fail.  I can revert the changes in src/{ffmpeg, libav} if you'd prefer to do this the right way.

This also brings a second commit along for the ride, which can be separated out if you'd prefer (but it depends upon the first commit due to nearby changes).  The second commit adds support for Colors.jl and makes checking for already-imported modules a little more conservative to prevent confusion between the old Color.jl and the new Colors.Color type.

This leaves only one deprecation warning: `FixedPointNumbers.Ufixed8`.  I'm not sure how to best resolve this since it's a "soft" dependency and we can't simply bump its version number in `REQUIRE`.


mbauman:
Bump.  I know this is a large diffset, but all changes were done with the Regex matches as described.  Is there any way I can make this easier for you to review?  Or get travis passing?

As it stands, VideoIO spits out so many deprecation warnings it's unusable in Jupyter notebooks.

kmsquire:
Sorry, @mbauman, I'll try to make time for this, but no guarantees.

I had fixed all of these a long time ago in the `refactor` branch.  It was close to ready, but my work on that stalled out months ago.  If you get the chance, can you try that branch?

PR #82, opened 2016-05-14
use dirname(@__FILE__) instead of Pkg.dir

tkelman:
allows installing the package elsewhere


tkelman:
looks like there are a bunch of these that all need to be fixed

PR #83, opened 2016-05-17
Added seek functions

gizle:
Hi

I needed to be able to seek in videos when using VideoIO, and added support for that. Since I don't know the conventions for this repo, and in general new to Julia, I will probably have violated some expected style. If you find something you don't like, please let me know.

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -1,6 +1,6 @@
 # AVIO
 
-import Base: read, read!, show, close, eof, isopen, seekstart
+import Base: read, read!, show, close, eof, isopen, seek, seekstart
 
 export read, read!, pump, openvideo, opencamera, playvideo, viewcam, play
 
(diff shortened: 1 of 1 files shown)


kmsquire:
Hi @gizle, thanks for the contribution.  Can you add some tests for this?

Ideally, we would like to know that the seek worked  Maybe something like

1. Find the FPS.
2. Read all frames up to 1 second, and store the last one.
3. Rewind, and then seek to 1 second.

Some of this is done in the tests already.  Could you modify the main loop to do something like this?

(If you aren't able to do this right now, even some non-checked tests of the functionality which simply call the new functions would be good.)

gizle:
I also see that it failed on Julia 0.3, that I did not test on at all. And I did not find the test suite. I will come back when I have fixed these issues.

gizle:
Should I just close this pull request and make a new one when I have a proper patch?

kmsquire:
Generally, you can just leave the pull requests open.  Any changes you push to the same branch will appear here.

However, I noticed that you're making changes from your own master branch.  This sometimes makes it hard to keep up with upstream changes, so you might consider creating a new branch, pushing that to your github repo, and reopening this PR from the new branch.

Tests are under `test/`.  Cheers!

PR #86, opened 2016-05-30
Adding seeking in video

gizle:
A new attempt to video seeking. This time tested on Julia 0.3, 0,4 and latest, and added tests. The tests did not run on Ubuntu 16.04 before I added my code, with the same problem as apparently was on OSX too, since the same video was conditionally excluded on OSX.  I have not fixed that, but can send a patch in a different pull request if required. The reason seems to be that 1080 is not divisible by 16, the macroblock size.

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -1,6 +1,6 @@
 # AVIO
 
-import Base: read, read!, show, close, eof, isopen, seekstart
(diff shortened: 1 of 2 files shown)


gizle:
Hmm. It still crash on travis, but works for me. Hard to debug then. I will see if I can imagine what goes wrong on travis,

kmsquire:
Closed by #102

PR #98, opened 2017-02-28
support ffmpeg 3.2.4

bjarthur:
closes https://github.com/kmsquire/VideoIO.jl/issues/76

just added "57" to the aliases, and reformatted for a narrower terminal.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -4,14 +4,28 @@ using BinDeps
 
 group = library_group("libav")
 
-deps = [
-    libavcodec = library_dependency("libavcodec", aliases=[["libavcodec-ffmpeg.so.","libavcodec.","libavcodec.so.","libavcodec.ffmpeg.so.","avcodec-"].*["53" "54" "55" "56"]...], group = group)
-    libavformat = library_dependency("libavformat", aliases=[["libavformat-ffmpeg.so.","libavformat.","libavformat.so.","libavformat.ffmpeg.so.","avformat-"].*["53" "54" "55" "56"]...], group = group)
-    libavutil = library_dependency("libavutil", aliases=[["libavutil-ffmpeg.so.", "libavutil.","libavutil.so.", "libavutil.ffmpeg.so.", "avutil-"].*["51" "52" "54"]...], group = group)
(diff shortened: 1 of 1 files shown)


kmsquire:
Thanks!  Just to make sure none of the individual package versions have been bumped, can you give the output of `VideoIO.versioninfo()`.

bjarthur:
```
julia> VideoIO.versioninfo()
Using ffmpeg
AVCodecs version 57.64.101
AVFormat version 57.56.101
AVUtil version 55.34.101
SWScale version 4.2.100
AVDevice version 57.1.100
AVFilters version 6.65.100
```

however...  the tests are crashing randomly:

```
julia(1620,0x7fffe94393c0) malloc: *** error for object 0x115540ff0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug

signal (6): Abort trap: 6
while loading /Users/arthurb/.julia/v0.5/VideoIO/test/avio.jl, in expression starting on line 20
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 5908834 (Pool: 5907800; Big: 1034); GC: 10
```

perhaps it is premature to support 3.2.4.

kmsquire:
I finally poked at this last weekend, and made some progress (though not a fix yet).

For convenience, since the first version, `AVFrame` objects have been allocated by VideoIO, rather than through the corresponding FFMpeg calls.  This has some nice benefits--automatic memory management, easy access to members.  

Unfortunately, starting with ffmpeg v3.1, some objects that manipulate `AVFrame` objects attempt to free/reallocate some of the memory, causing the problems noted above.

I have a version locally which fixes this error (by allocating frames with ffmpeg--will push the branch soon).  However, it currently breaks the tests, and `seek` doesn't work.  So it's not quite ready yet.

That said, I think it's close, and it will be nice to be able to use a recent version of ffmpeg (especially for Windows users).

ianshmean:
I believe this is corrected/outdated now, given newer ffmpeg versions are supported, so closing. Feel free to open if that's not the case and this specific version is still needed

PR #102, opened 2017-07-27
Another attempt at a seek functionality

romainFr:


romainFr:
hmm, av_frame_get_best_effort_timestamp is ffmpeg specific, hence the failure on Travis (which uses libav if I understand correctly)... will need to find a workaround.

romainFr:
It seems to be working now. Travis fails on 0.7, but it's due to other packages failing to precompile there. Should we change the Travis config file to allow failures on nightly ?

kmsquire:
Sweet!  Yes, updating Travis confit to allow failures on 0.7 would be good. Should probably be a separate PR, but here is fine if you want.

kmsquire:
V0.6 should also be added.

romainFr:
I made a separate PR for the Travis config.

romainFr:
I'm also just realizing that this PR removes ImageCore and ImageView from REQUIRE. Should I revert that ?

kmsquire:
> I'm also just realizing that this PR removes ImageCore and ImageView from REQUIRE. Should I revert that ?

Another thing that probably should be done (it prevents precompile), but it should probably be in a separate PR.  (I think the README also refers to some of the functionality that would be removed.)

romainFr:
I think this is clean now. I removed all changes unrelated to the seek function and added a mention in the readme. Anything else ?

src/avio.jl
```diff
@@ -1,10 +1,10 @@
 # AVIO
 
-import Base: read, read!, show, close, eof, isopen, seekstart
+import Base: read, read!, show, close, eof, isopen, seek, seekstart
 
 export read, read!, pump, openvideo, opencamera, playvideo, viewcam, play
 
-using Compat
+using Compat, FileIO
```
kmsquire: I'm curious what you're using from `FileIO`?  Is this the `imwrite` => `save` change in `test/avio.jl`?  If so, then   this import only needs to be in the test file, and not here.  Or was something broken before and fixed by this statement?

src/avio.jl
```diff
@@ -501,19 +466,51 @@ have_frame(r::StreamContext) = !isempty(r.frame_queue) || have_decoded_frame(r)
 have_frame(avin::AVInput) = any(Bool[have_frame(avin.stream_contexts[i+1]) for i in avin.listening])
 
 reset_frame_flag!(r) = (r.aFrameFinished[1] = 0)
+function seconds_to_timestamp(s::Float64, time_base::AVRational)
```
kmsquire: A blank line before this function would be nice.

src/avio.jl
romainFr: No, you're right, it's useless here, I removed it...

src/avio.jl
romainFr: True

kmsquire:
Thanks!

PR #104, opened 2017-07-29
Adding appveyor.yml and appveyor badge

romainFr:
Very similar to previous similar PRs. I added 0.5, 0.6 and nightly (failures being allowed there).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-[![Build Status](https://travis-ci.org/kmsquire/VideoIO.jl.svg?branch=master)](https://travis-ci.org/kmsquire/VideoIO.jl) [![Coverage Status](https://coveralls.io/repos/kmsquire/VideoIO.jl/badge.png)](https://coveralls.io/r/kmsquire/VideoIO.jl)
+[![Build Status](https://travis-ci.org/kmsquire/VideoIO.jl.svg?branch=master)](https://travis-ci.org/kmsquire/VideoIO.jl)
+[![AppVeyorBuild status](https://ci.appveyor.com/api/projects/status/github/kmsquire/VideoIO.jl?svg=true&branch=master)](https://ci.appveyor.com/project/kmsquire/videoio-jl/branch/master)
+[![Coverage Status](https://coveralls.io/repos/kmsquire/VideoIO.jl/badge.png)](https://coveralls.io/r/kmsquire/VideoIO.jl)
 
 VideoIO.jl
 ==========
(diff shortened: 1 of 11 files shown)


romainFr:
For now it fails to build on appveyor because the oldest binary ffmpeg provides is 3.2 (which is libavcodec 57). Should we use an unofficial source for the old binaries, or wait for #98 to be a viable solution ?

kmsquire:
@romainFr, thanks for working on this!

Short answer (and short term solution): If you know of an unofficial source for the old binaries, we could go with those for now.

Longer answer: when I initially created this package, I had the ambition to support many versions of libav and ffmpeg, as well as as much of the interface as seemed reasonable.  To this end, I used Clang.jl to wrap the libraries, creating function wrappers for almost everything, but only exposing the functions needed to get things to run.

This turned out to be a little too ambitious.  The Clang wrapper gets everything wrapped, but despite a lot of work, still requires a bit of manual fixing of the interfaces.  It also has required updating as Julia and Clang.jl have been updated.  Changes in my own circumstances have restricted the amount of time I've had to spend on Julia development in general, and this package in particular, and so it's basically been limping along as others (like you) have helped fix it up.

If I were to do it again, I would probably focus on the few types, functions, and constants that are actually needed to have a minimal video reader/writer running, and drop the rest.

Now that I'm here, I'll see what I can do tonight to help here (probably by trying to wrap later versions of ffmpeg), but I probably won't have much time again anytime soon after this.

Since you're interested and using the package, would you be interested in either taking over maintenance, or becoming a contributor with write permission here?

kmsquire:
(I could and should also move this over to the JuliaImages organization...)

romainFr:
@kmsquire thanks ! I'd be happy to be a contributor, not entirely sure  how much time I'll manage to save for it but I'm glad to help. As to moving VideoIO I was thinking of the following direction : 
- move the input/output part to JuliaIO. ```Load``` could return either an array or a ```VideoReader``` depending on how many frames the user requests
- move the display part of the package, either to a VideoView package or as a part of [ImageView.jl](https://github.com/timholy/ImageView.jl), as a ```VideoReader``` method to ```imshow``` Maybe @timholy has something to say about that.

kmsquire:
> I'd be happy to be a contributor, not entirely sure how much time I'll manage to save for it but I'm glad to help. 

Done!

> As to moving VideoIO I was thinking of the following direction :
> 
> * move the input/output part to JuliaIO. Load could return either an array or a VideoReader depending on how many frames the user requests
> * move the display part of the package, either to a VideoView package or as a part of ImageView.jl, as a VideoReader method to imshow Maybe @timholy has something to say about that.

That sounds perfect--I had something quite similar in mind.  If you have time to work toward any of these, that would be much appreciated!

(Didn't actually get the wrapping of later ffmpeg/libav versions done, but I'll try to work on that this week.)

ianshmean:
Closing as covered by #138 and a few of the things discussed have been enacted

PR #113, opened 2017-12-10
Fix v0.6 deprecations and drop v0.5 support

rdeits:
Thanks for making this project! I just noticed that it has a few deprecation warnings in v0.6, so I fixed the warnings. I also dropped v0.5 support, since it's almost time for v0.7/1.0, and dropping 0.5 will let FemtoCleaner magically help with some of the syntax changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ addons:
     - libcairo2
     - libavcodec-extra-54
     - libavdevice53
-    - libavfilter2
-    - libavformat53
-    - libavutil-extra-51
+    - libavfilter3
+    - libavformat54
+    - libavutil-extra-52
     - libswscale2
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ os:
   - linux
 #  - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 matrix:	
@@ -30,5 +29,4 @@ notifications:
 script:
   - julia -e 'Pkg.clone(pwd())'
   - julia -e 'Pkg.build("VideoIO")'
-  # coverage=true causes tests to hang, see pull request #92
-  - if [ -f test/runtests.jl ]; then julia -e 'Pkg.test("VideoIO", coverage=false)'; fi
+  - if [ -f test/runtests.jl ]; then julia -e 'Pkg.test("VideoIO", coverage=true)'; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - tcl8.5
     - tk8.5
     - libcairo2
-    - libavcodec-extra-53
+    - libavcodec-extra-54
     - libavdevice53
     - libavfilter2
     - libavformat53
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ video frames from a supported video file, or from a camera device:
 
     # One can seek to an arbitrary position in the video
     seek(f,2.5)  ## The second parameter is the time in seconds and must be Float64
-    img = read(f, Image)
+    img = read(f)
     canvas, _ = ImageView.view(img)
     
     while !eof(f)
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -30,5 +29,4 @@ notifications:
 script:
   - julia -e 'Pkg.clone(pwd())'
   - julia -e 'Pkg.build("VideoIO")'
-  # coverage=true causes tests to hang, see pull request #92
-  - if [ -f test/runtests.jl ]; then julia -e 'Pkg.test("VideoIO", coverage=false)'; fi
+  - if [ -f test/runtests.jl ]; then julia -e 'Pkg.test("VideoIO", coverage=true)'; fi
```
rdeits: I've had issues with `coverage=true` on travis, but only on v0.5. We'll see if changing this back causes any issues on v0.6.

kmsquire:
Thanks for the contribution, and glad you're finding it useful!

Unfortunately, I haven't had time to update it to finish updating it to more recent versions of ffmpeg/libav yet, which is why CI is failing. 

That pull request is here (but still needs some work, and I'm not sure when I'll get to it...): https://github.com/kmsquire/VideoIO.jl/pull/108

kmsquire:
Anyway I assume this works for you locally?  If so, I'll merge.

rdeits:
It does pass for me locally, but I'm on OSX, so I'm running a very different set of binary dependencies than Travis. I did manage to get Travis past the `apt-get install` phase, but it's now throwing what looks like it might be a real error:

```
ERROR: LoadError: LoadError: InexactError()

Stacktrace:

 [1] convert(::Type{Int64}, ::Float64) at ./float.jl:679

 [2] seek(::VideoIO.VideoReader{true}, ::Float64, ::Float64, ::Float64, ::Int64, ::Bool) at /home/travis/.julia/v0.6/VideoIO/src/avio.jl:490

 [3] macro expansion at /home/travis/.julia/v0.6/VideoIO/test/avio.jl:55 [inlined]

 [4] anonymous at ./<missing>:?

 [5] include_from_node1(::String) at ./loading.jl:576

 [6] include(::String) at ./sysimg.jl:14

 [7] include_from_node1(::String) at ./loading.jl:576

 [8] include(::String) at ./sysimg.jl:14

 [9] process_options(::Base.JLOptions) at ./client.jl:305

 [10] _start() at ./client.jl:371

while loading /home/travis/.julia/v0.6/VideoIO/test/avio.jl, in expression starting on line 20

while loading /home/travis/.julia/v0.6/VideoIO/test/runtests.jl, in expression starting on line 1
```

The fix would be to change that `convert(Int` to a `round(Int`, but I don't know enough about this package to be sure that that's the right thing to do.

kmsquire:
> The fix would be to change that `convert(Int` to a `round(Int`, but I don't know enough about this package to be sure that that's the right thing to do.

Just looked--this is the correct fix.

rdeits:
Something else is going on: changing the `convert` to a `round` still gives an InexactError, which means the argument is probably `Inf` or `NaN`. I added a debugging statement to see if I can figure it out (and I'll try to test this locally on linux to avoid making Travis do all the work)

kmsquire:
Thanks for putting in the time!

rdeits:
Ok, the actual problem is that for `annie_oakley.ogg`, the frame rate's numerator is zero: 

```
Test Failed

  Expression: v.stream_info.stream.r_frame_rate.num != 0

   Evaluated: 0 != 0
```

of course, I don't know why *that* happened yet, but at least it's a step in the right direction.

rdeits:
Ok, the actual problem is that for `annie_oakley.ogg`, the frame rate's numerator is zero: 

```
Test Failed

  Expression: v.stream_info.stream.r_frame_rate.num != 0

   Evaluated: 0 != 0
```

of course, I don't know why *that* happened yet, but at least it's a step in the right direction.

rdeits:
I can reproduce the error in Ubuntu 14.04 (which is what Travis uses) with this docker file: https://gist.github.com/rdeits/cf839be3bfc1be386a883b47cf8eff9f

The same configuration seems to work fine for Ubuntu 16.04.

PR #117, opened 2018-03-25
Fixed example to be up to date

mbeltagy:
with changes in ImageView.jl

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,12 +69,12 @@ video frames from a supported video file, or from a camera device:
 
     # One can seek to an arbitrary position in the video
     seek(f,2.5)  ## The second parameter is the time in seconds and must be Float64
-    img = read(f, Image)
-    canvas, _ = ImageView.view(img)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,12 +69,12 @@ video frames from a supported video file, or from a camera device:
 
     # One can seek to an arbitrary position in the video
     seek(f,2.5)  ## The second parameter is the time in seconds and must be Float64
-    img = read(f, Image)
-    canvas, _ = ImageView.view(img)
+    img = read(f)
+    canvas=ImageView.imshow(img)["gui"]["canvas"]
```
kmsquire: Would you mind adding spaces around the `=`, to match the surrounding code convention?

mbeltagy:
I think we could just leave as it is. On reflection, I do agree that it it very likely that the user would want to do something with `Images`. The example code would offer a nice template for future exploration.

PR #124, opened 2018-08-21
Drop v0.6 support

kmsquire:


kmsquire:
Current failure is https://github.com/JuliaIO/ImageMagick.jl/issues/130.

PR #114, opened 2018-02-04
WIP: Turn on precompilation

timholy:
This is a not-yet-working attempt to use precompilation. It switches to using internal submodules rather than modifying `LOAD_PATH`, and (for now) it simply comments out the code that's conditional on ImageView being loaded. Before merging this we should decide on a more "serious" solution to this issue, such as:
- create a WebCam.jl package that depends on both VideoIO and ImageView
- have VideoIO return objects that ImageView can handle. For example, `opencamera` (or some higher-level wrapper) could conceivable return a [`StreamingContainer`](https://github.com/JuliaImages/ImageAxes.jl/blob/3e9d83e7c5d15184616e9b6ae6129e5402a38385/src/ImageAxes.jl#L193-L225) or something similar. Likewise, now that we have `seek` functionality, perhaps we could create a `VideoArray<:AbstractArray` that has the standard AbstractArray interface and lets ImageView (and any other Julia code) treat the movie file as if it's a giant array.

However, there's a more immediate problem: the tests don't pass. With precompilation, I get this:
```julia
julia> include("runtests.jl")
WARNING: Method definition zero(Type{VideoIO.AVUtil.AVRational}) in module AVUtil at /home/tim/.julia/v0.6/VideoIO/src/ffmpeg/AVUtil/src/../../../ffmpeg/AVUtil/v54/libavutil_h.jl:691 overwritten at /home/tim/.julia/v0.6/VideoIO/src/ffmpeg/AVUtil/src/AVUtil.jl:7.
VideoFile:
   name:         annie_oakley.ogg  (Downloaded)
   description:  The "Little Sure Shot" of the "Wild West," exhibition of rifle shooting at glass balls.
   license:      pubic_domain (US)
   credit:       
   source:       http://commons.wikimedia.org/wiki/File:Annie_Oakley_shooting_glass_balls,_1894.ogg
   download_url: https://upload.wikimedia.org/wikipedia/commons/8/87/Annie_Oakley_shooting_glass_balls%2C_1894.ogv
 
WARNING: deprecated syntax "abstract StreamContext".
Use "abstract type StreamContext end" instead.

VideoFile:
   name:         crescent-moon.ogv  (Downloaded)
   description:  Moonset (time-lapse).
   license:      Creative Commons Attribution 2.0 Generic (http://creativecommons.org/licenses/by/2.0/deed)
   credit:       Photo : Thomas Bresson
   source:       http://commons.wikimedia.org/wiki/File:2010-10-10-Lune.ogv
[...]

PR #108, opened 2017-09-09
WIP: Fix ffmpeg 3.1+

kmsquire:
This has been broken for a while.  I finally have a fix, although the code here isn't complete yet.

PR #125, opened 2018-08-22
don't modify LOAD_PATH, and fix some deprecations on 0.7

rdeits:
@kmsquire this is a PR into your branch from #124 

This isn't all the way done, but it's the best I can do right now, and I hope it's helpful. I fixed a few issues with 0.7+, but a few more remain:

### Load path modification

The load path modifications weren't working for me on 0.7, so I changed all of the conditional import stuff to just use `include()` as usual. I think this is an improvement anyway, as the previous behavior was dangerous: if a user happened to create a package or module called `AVUtil`, then Julia might load their AVUtil or VideoIOs, depending on the load order, causing both packages to be broken. Using `include()` avoids the issue entirely. 

However, I noticed in the Readme a note that not all of the submodules are actually loaded and a suggestion that users can load them as they see fit. That's no longer true after this change. But with precompilation, there should be no need to avoid just `include()`-ing everything. Precompilation is currently turned off for this module, so I guess the next thing to do would be to (1) move all the run-time initialization to the `__init__()` function, (2) enable precompilation (3) actually `include()` all the submodules no matter what. That should fix the issue completely and also make the package easier and more reliable to work with. 

### Conditional loading

Packages aren't loaded into `Main` in v0.7 and above, so trying to inspect `Main.ImageView` doesn't work. Fortunately https://github.com/MikeInnes/Requires.jl solves the actual issue or conditionally loading glue code, and also removes the need to load ImageView before loading VideoIO. Note that my change doesn't include the fallback `play()` methods that just print error messages, but those could be added back in. 

### Remaining deprecations

There are a few more deprecations, but they're repeated across all of the many many mostly-duplicate files and I don't have time to go through them all, sorry...

### Status

With this PR, I can at least load the package on v0.7. Testing gets as far as reading `annie_oakley.ogg` before it gets aborted with a `double free or corruption` on my Ubuntu 18.04 system with: 

```julia
julia> VideoIO.versioninfo()
Using ffmpeg
AVCodecs version 57.107.100
AVFormat version 57.83.100
AVUtil version 55.78.100
SWScale version 4.8.100
AVDevice version 57.10.100
AVFilters version 6.107.100
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 VideoIO.jl
 ==========
 
-Julia bindings for libav/ffmpeg.  
+Julia bindings for libav/ffmpeg.
 
-Currently, only video reading is supported, for the following 
+Currently, only video reading is supported, for the following
 library versions:
 
 * libav 0.8, 9, 10
@@ -14,9 +14,9 @@ library versions:
 Video images may be read as raw arrays, or optionally, `Image`
 objects (if `Images.jl` is installed and loaded first).
 
-Feel free to request support for additional libav/ffmpeg 
+Feel free to request support for additional libav/ffmpeg
 versions, although earlier versions may be too challenging to
-support. 
+support.
 
 If you encounter any problems, please add the output
 of `VideoIO.versioninfo()` to your report.
@@ -41,13 +41,10 @@ A trivial video player interface exists (no audio):
     # Aternatively, you can just open the camera
     #VideoIO.viewcam()
 
-Note that `ImageView` must be imported before `VideoIO` for the `playvideo`
-function to be available.
-
 High Level Interface
 --------------------
 
-VideoIO contains a simple high-level interface which allows reading of 
+VideoIO contains a simple high-level interface which allows reading of
 video frames from a supported video file, or from a camera device:
 
     using Images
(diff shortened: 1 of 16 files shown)


kmsquire:
@rdeits, thank you again!  I've sent an invitation to add you as a collaborator here.

kmsquire:
With respect to pre-compilation and conditional loading: we've discussed previously about creating a WebCam.jl (or something similar), and then remove the dependency on ImageView.  See @timholy's work in #114.  

For the `double free or corruption` issue, I think that's probably related to #108.  

Around FFMpeg v3.1, some things changed internally in how FFmpeg manages memory that made the code here break.  
When I first made the package, I tried hard to allocate all memory in Julia, so it could be managed by the garbage collector.  
But ffmpeg really wants to do memory management itself, so the way we create and use `AVFrames` and `AVPackets` 
(and probably a few other things) needs to change. 

#108 has been my attempts to do that, but it doesn't quite work yet.  I (or someone else?) need to create a debug build of ffmpeg and probably julia, 
and see if I can trace into it to find the problem.  If you happen to have time to look at that, that would be great.  (#108 needs to be rebased)

PR #138, opened 2019-02-08
Updating windows to ffmpeg 4.1 and link all handled lib versions

ianshmean:
Currently build completes without error, but `] test VideoIO` fails because ffmpeg isn't present in the VideoIO deps folder

PR #140, opened 2019-02-12
Relax `video_stream` type in `seek`

kmsquire:
* This seems to be causing failures on 32-bit Windows, at least


kmsquire:
Encapulated by #138.

Issue #60, opened 2015-07-09
Can we make VideoIO an IO only library!?


SimonDanisch:
Hi,
as I'm pretty much implementing the ImageView functionality over at GLVisualize, it's pretty annoying that I suddenly depend on Images, ImageView, TK, Cairo, etc, just by using VideoIO.
Would it make sense to put this functionality into some package like e.g. VideoView?!

Best,
Simon

kmsquire:
Hi Simon, I'd be open to something like that. Unfortunately, I probably won't be able to work on it for the next week or so.

timholy:
If you're effectively rewriting ImageView, I should share my notes on a planned-someday-when-I-have-time revamp. Not sure any of this will make sense (maybe even to me!), but here's the file I found on my laptop:

```
ImageView revamp notes:

- New view types: IVArray{T,N,(:X,:Y)}, IVArray{T,N,(:Y,:X)}, IVArray{T,N,(:X,:Y,:T)}, IVArray{T,N,(:Color,...)}.
  This should end all the need for slicing
- Don't allocate a new surface with each resize, just use the old one and a subset if necessary. Access the subset by BoundingBox
- Use traits. has_time() and size_time_known() (to include streaming VideoIO objects of unknown or infinite duration), has_pixels() (for function plotting, like fractals with infinite zoomability)
- Probably use Reactive for controls
- Think about state & containment more carefully. Support current usage, two panels displaying two images locked to a single set of controls, a toggle viewer, a display like stackviewer, a 3x3 grid with different z slices and a single t control.
- Or better not to use SubArrays in ImageView, and instead use a CartesianRange for specifying location?

Look at Holoviews for inspiration, http://ioam.github.io/holoviews.
```

SimonDanisch:
Holoview looks interesting =)
Could be done with GLVisualize reeelatively easily... I just need to be carefull with hacking in new futures which don't generalize well.
Where do you use slicing so far? Or is that for volume images?
Reactive is actually surprisingly well fitted for video streams. I'm using a Reactive signal of images for displaying videos and its completely smooth ;) With some additional work we could paramertrice the signal more, to allow for pause, fast forward/backward and stepping to specific times.
How would I currently jump to a specific time point in a video stream!?

timholy:
Yes, slicing is for volume images. And as you say, time-jumping only works for something for which you have random access. Both of which are true for data sets I care a lot about, but generalization is a good thing.

[...]

PR #141, opened 2019-02-15
WIP: Move to BinaryProvider

ianshmean:
Attempting to make the build process more robust by moving to BinaryProvider.jl


deps/build.jl
```diff
@@ -1,133 +1,61 @@
...
+    LibraryProduct(libpath, [["libavutil-ffmpeg.so.", "libavutil.","libavutil.so.", "libavutil.ffmpeg.so.", "avutil-"].*["51" "52" "54" "55" "56"]...], :libavutil),
+    LibraryProduct(libpath, [["libswscale-ffmpeg.so.","libswscale.","libswscale.so.","libswscale.ffmpeg.so.","swscale-"].*["2" "3" "4" "5"]...], :libswscale),
+    LibraryProduct(libpath, [["libavfilter-ffmpeg.so.","libavfilter.","libavfilter.so.","libavfilter.ffmpeg.so.","avfilter-"].*["2" "3" "4" "5" "6" "7"]...], :libavfilter),
+    LibraryProduct(libpath, [["libavdevice-ffmpeg.so.","libavdevice.","libavdevice.so.","libavdevice.ffmpeg.so.","avdevice-"].*["53" "54" "55" "56" "57" "58"]...], :libavdevice),
+]
+
+# Download binaries from hosted location
+v = "4.1"
+#bin_prefix = "https://github.com/ianshmean/FFMPEG-tarballs/raw/master/$v/bin"
+bin_prefix = "/Users/IanB/Github_Personal/FFMPEG-tarballs/$v/bin"
```
staticfloat: You probably meant to keep this as an http URL.  ;)

deps/build.jl
```diff
@@ -1,133 +1,61 @@
...
+    ExecutableProduct(prefix, "ffprobe", :ffprobe),
+    ExecutableProduct(prefix, "ffplay", :ffplay),
+
+    LibraryProduct(libpath, [["libavcodec-ffmpeg.so.","libavcodec.","libavcodec.so.","libavcodec.ffmpeg.so.","avcodec-"].*["53" "54" "55" "56" "57" "58"].*[".dylib"]...], :libavcodec),
+    LibraryProduct(libpath, [["libavformat-ffmpeg.so.","libavformat.","libavformat.so.","libavformat.ffmpeg.so.","avformat-"].*["53" "54" "55" "56" "57" "58"]...], :libavformat),
+    LibraryProduct(libpath, [["libavutil-ffmpeg.so.", "libavutil.","libavutil.so.", "libavutil.ffmpeg.so.", "avutil-"].*["51" "52" "54" "55" "56"]...], :libavutil),
+    LibraryProduct(libpath, [["libswscale-ffmpeg.so.","libswscale.","libswscale.so.","libswscale.ffmpeg.so.","swscale-"].*["2" "3" "4" "5"]...], :libswscale),
+    LibraryProduct(libpath, [["libavfilter-ffmpeg.so.","libavfilter.","libavfilter.so.","libavfilter.ffmpeg.so.","avfilter-"].*["2" "3" "4" "5" "6" "7"]...], :libavfilter),
+    LibraryProduct(libpath, [["libavdevice-ffmpeg.so.","libavdevice.","libavdevice.so.","libavdevice.ffmpeg.so.","avdevice-"].*["53" "54" "55" "56" "57" "58"]...], :libavdevice),
+]
```
staticfloat: These `LibraryProduct` definitions are probably more complex than they need to be.  Do you actually care that the versions are one of `53:58`, or can it just be any version number?  Ideally, you'd just be giving names like `["libavcodec", "libavcodec.ffmpeg"]` and BinaryProvider will take care of dealing with `.53.so`, or `.so`, or `.dylib`, or whatever.  If there is a definitive list of what these things can be called, I can help you make the simplest possible `LibraryProduct` line.

src/testvideos.jl
```diff
@@ -36,15 +36,15 @@ const videofiles  =  Dict(
...
+                                                "http://archive.org/download/LadybirdOpeningWingsCCBYNatureClip/Ladybird%20opening%20wings%20CC-BY%20NatureClip.mp4",
                                                 ),
 
                     "annie_oakley.ogg" => VideoFile("annie_oakley.ogg",
                                                     "The \"Little Sure Shot\" of the \"Wild West,\" exhibition of rifle shooting at glass balls.",
                                                     "pubic_domain (US)",
                                                     "",
                                                     "http://commons.wikimedia.org/wiki/File:Annie_Oakley_shooting_glass_balls,_1894.ogg",
-                                                    "https://upload.wikimedia.org/wikipedia/commons/8/87/Annie_Oakley_shooting_glass_balls%2C_1894.ogv"),
+                                                    "http://upload.wikimedia.org/wikipedia/commons/8/87/Annie_Oakley_shooting_glass_balls%2C_1894.ogv"),
```
staticfloat: :(

deps/build.jl
ianshmean: Indeed! I was working on this on a spotty internet connection so had gone local and forgot to revert. just fixed it. thanks

src/testvideos.jl
ianshmean: Same as above, for some reason my spotty internet connection (on a united flight) wasn't liking `download()`ing from https urls. I just changed them back

deps/build.jl
ianshmean: This format/list is inherited from the BinDeps setup, and the whole package is built out for specific library versions given the degree of changes that ffmpeg has had since the earlier supported versions, so I believe we do need to be specific, but perhaps ranges would suffice.

staticfloat: Since you have the actual URLs below, I bet we can cut this down to just the versions within the tarballs and make this a lot simpler.

ianshmean: The problem is that I haven't yet found linux platform-specific shared binaries, which would be ideal to control version like we're trying in windows and MacOS. 
Given linux typically(?) comes preinstalled with ffmpeg, the coverage of all the lib versions is needed for that. Perhaps there's an alternative that I'm missing..

staticfloat:
Switching to BinaryProvider like this is not going to allow you to use
preinstalled binaries. With BP, (unlike BinDeps) by default you can only
use the binaries that are pointed to by the URLs in the build script. You
would have to jump through a lot of hoops to get it to pick up the binaries
from the environment, and that is by design.

ianshmean:
Ok, that makes sense. 
This branch is now working on Windows and MacOS, and should be ready for the linux binaries, so we can just leave it until the time is right.

ianshmean:
I just tried turning precompilation on for this and it shows the same error during testing as https://github.com/JuliaIO/VideoIO.jl/pull/114 where `Testing IO reading...` fails when attempting to use `avformat_open_input`

I've found some discussion on `avformat_open_input` crashing but perhaps someone more familiar with ffmpeg than me could advise? @kmsquire ?
https://stackoverflow.com/questions/26552897/avformat-open-input-function-crashing

rdeits:
Does precompilation work for this branch now that #143 has been merged? 

Also, could we potentially get the linux FFMPEG binaries from https://github.com/jpsamaroo/FFMPEGBuilder (potentially with some additional updates to that repo)?

ianshmean:
@rdeits Indeed, precompillation is working in this branch now.

And I like the idea of grabbing the linux builds if they're within reach from FFMPEGBuilder.. I'm sure others are much more familiar than me with the issues around doing that than me

ianshmean:
Just to summarise. This PR is waiting for Linux ffmpeg binaries. I've not had time to create them yet, or explore getting them from https://github.com/jpsamaroo/FFMPEGBuilder

If anyone does, it would be great to merge this

jpsamaroo:
I'm happy to transfer that repo wherever if people want to work on it; I don't have the bandwidth or desire to figure out the YASM build issues right now.

SimonDanisch:
@jpsamaroo, can you move the builders to JuliaIO?

ianshmean:
How about this as an interim solution, while the FFMPEG linux binaries aren't ready.. we build with Bindeps if Linux, and BinaryProvider if Windows or MacOS?

ianshmean:
I've implemented an approach of BinDeps for Windows, BinaryProvider for everything else.

Side note.. Travis is passing but with a segfault during Cairo build on MacOS due to the ongoing Cairo/Homebrew issue https://github.com/JuliaPackaging/Homebrew.jl/issues/253 (which is in itself an example of why #60 is a good idea given `Cairo` is a dep through `ImageView`)

ianshmean:
Scratch that last comment, I finally got my head around FFMPEGbuilder via Travis (simpler than I expected) and got a fork releasing the linux binary (https://github.com/ianshmean/FFMPEGBuilder) so this PR should now support GNU linux, MacOS and Windows, if all tests pass.

jpsamaroo:
In order for your build of FFMPEG to be decent, the rest of the deps will also need to support all the same platforms, so their is probably still lots of work to do (unless you updated all of them without me noticing :) ).

ianshmean:
@jpsamaroo I just plucked out the GNU linux binary from FFMPEGbuilder, the windows and MacOS binaries came from: https://ffmpeg.zeranoe.com/builds/

Saying that, looking at Travis, the linux binary libs don't seem to satisfy BinaryProvider, which I guess means they're not dlopen-able, given they are present in the dir

SimonDanisch:
Seems like `libfdk-aac` is missing, which leads to `libavdevice` to error, which makes it unsatisfied :(

SimonDanisch:
Seems like it's already included (`--enable-libfdk-aac --enable-nonfree`):
```julia
  --enable-nonfree     \
  --disable-static     \
  --enable-shared      \
  --disable-debug      \
  --enable-avresample  \
  --enable-libass      \
  --enable-libfdk-aac  \
```
 Do we actually need a library product for it, to not "loose it", or is there something else going wrong?

SimonDanisch:
Ok, this should hopefully fix it :)

ianshmean:
Ah! @SimonDanisch I was rebasing too! 😂

SimonDanisch:
Of course, I knew this would happen :D

ianshmean:
You do what's easiest. Especially if you're on a roll with getting this working!

SimonDanisch:
continued in https://github.com/JuliaIO/VideoIO.jl/pull/159

PR #143, opened 2019-02-23
Enable precompilation(!!!)

kmsquire:
The test (and code) that was failing for precompilation was code that reads from a Julia `IOStream`.  Reading inputs directly with ffmpeg's own I/O functions was working fine.

In order to read from a Julia `IOStream`, we pass a callback function which fills a pre-allocated buffer with data from the stream.  The pointer to that callback has to be determined at runtime (in `__init__`), because the function location will change each time the program is run.

Cc: @timholy @ianshmean

### DIFF
--- a/src/VideoIO.jl
+++ b/src/VideoIO.jl
@@ -1,5 +1,3 @@
-__precompile__(false)
-
 module VideoIO
 
 using FixedPointNumbers, ColorTypes, ImageCore, Requires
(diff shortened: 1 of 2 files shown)


ianshmean:
Fantastic! And I see that CI tests are passing on all platforms!! Nice one! 🎉

kmsquire:
CI all passed--going to merge.

I'm slightly curious if the fact that `read_packet` is global and not constant (because it's initialized in `__init__`) causes much more work for the compiler because it can't know if the type will change.  

I thought about moving the `@cfunction` initialization to the `open_avinput` function, but then I believe the variable would go out of scope and be garbage collected.  There's probably (?) an easy way around this, so if anyone (@timholy) has a better suggestion, I'd be all ears.

SimonDanisch:
use `const func = Ref{Ptr{Cvoid}}()`?

kmsquire:
Thanks @SimonDanisch!

kmsquire:
#144

PR #110, opened 2017-09-14
Add and test get_duration() get_start_time() get_time_duration()

yakir12:
fixes #99 
This builds on #109, so use this PR, not #109...

PR #153, opened 2019-04-22
Add and test get_duration() get_start_time() get_time_duration() (v2 of #110)

ianshmean:
Updated version of #110


yakir12:
This is awesome! Thank you.

PR #159, opened 2019-04-29
use binaryprovider

SimonDanisch:
Rebased #141


SimonDanisch:
woah, all green! What a time to be alive^^
I'd say we try to squash this a bit before merging

ianshmean:
Fantastic! Squashing sounds good! Especially given the mess I made just updating the readme a little..

SimonDanisch:
Uh, did you just undo my squashing with your last commit?! :D

SimonDanisch:
Should I force push my squashed history again, and you reapply your last commit without getting all the other commits back?

ianshmean:
Ah! Sorry... Don't know how that happened. Yeah. Feel free to modify the readme if you want to do it at the same time, or happy to after you

SimonDanisch:
This is ready to merge in my opinion :)

ianshmean:
I think so too. @kmsquire what do you think? If we did, I think we should bump version to v0.6.0

deps/build.jl
```diff
@@ -1,133 +1,95 @@
...
+# Download binaries from hosted location
+v = "4.1"
+bin_prefix = "https://github.com/ianshmean/FFMPEG-tarballs/raw/master/$v/bin"
+
+download_info = Dict(
+    #Linux(:aarch64, :glibc) => ("$bin_prefix/ffmpeg-$v-aarch64-linux-gnu.tar.gz", ""),
+    #Linux(:armv7l, :glibc)  => ("$bin_prefix/ffmpeg-$v-arm-linux-gnueabihf.tar.gz", ""),
+    #Linux(:powerpc64le, :glibc) => ("$bin_prefix/ffmpeg-$v-powerpc64le-linux-gnu.tar.gz", ""),
+    #Linux(:i686, :glibc)    => ("$bin_prefix/ffmpeg-$v-i686-linux-gnu.tar.gz", ""),
+    Linux(:x86_64, :glibc)  => ("$bin_prefix/ffmpeg-4.1-86_64-linux-gnu.tar.gz", "7d1199e95f8a81b6e3fa6fd6584ca6f419847f6d4f5a5dc83656cd46e84a51fb"),
```
kmsquire:
```suggestion
    Linux(:x86_64, :glibc)  => ("$bin_prefix/ffmpeg-$v-86_64-linux-gnu.tar.gz", "7d1199e95f8a81b6e3fa6fd6584ca6f419847f6d4f5a5dc83656cd46e84a51fb"),
```

deps/build.jl
```diff
@@ -1,133 +1,95 @@
...
+        Mod = @eval module Anon end
+        Mod.include(file)
+    end
+    # Not sure why those end up in lib, but they do -.-
+
+    for elem in readdir(joinpath(prefix, "lib"))
+        if occursin(".so", elem)
+            cp(joinpath(prefix, "lib", elem), joinpath(prefix, "bin", elem), follow_symlinks = true, force = true)
+        end
+    end
```
kmsquire: I was actually curious why the libraries ended up in `bin`, rather than `lib`... @SimonDanisch, is there a reason why `BinaryProvider` puts everything in `bin`, rather than in `lib`?

ianshmean: Actually, that's my doing.. the ffmpeg shared tarballs which are used for Windows and MacOS are from [here](https://ffmpeg.zeranoe.com/builds/), and they bundle everything into `bin` so I modified the `BinaryProvider` tarball to have everything in `bin` too.

Is the current situation ok, or should I edit the tarballs? It would be good to start off on the right foot!

deps/build.jl
ianshmean: Good catch!

deps/build.jl
ianshmean: I've decided to leave it as-is. Next time we release a new FFMPEG version we can conform to the `bin` and `lib` standard

ianshmean:
Thanks so much @SimonDanisch!

PR #160, opened 2019-05-02
Replace ImageView video playback with Makie

ianshmean:
The current `ImageView` method seems to be broken, and this fixes it (for me at least).

It would be great to add play/pause, track, progress indicator, frame number, time etc, if those are easy to add now?

```julia
julia> using Makie, VideoIO
julia> VideoIO.playvideo("localpath/VideoIO.jl/videos/annie_oakley.ogg")
```
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Simple Interface
 ----------------
 A trivial video player interface exists (no audio):
 
-    import ImageView
+    import Makie
     import VideoIO
 
     f = VideoIO.testvideo("annie_oakley")  # downloaded if not available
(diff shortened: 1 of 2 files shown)


timholy:
Given the GTK problems, I'm all in favor of this. If I can convince myself that Makie works cross-platform and on remote machines, I might switch ImageView to use it.

asinghvi17:
@timholy You'll need a graphics card which supports OpenGL v3.3 at least to run Makie.  It is possible but painful to get it to run using software OpenGL rendering.

PR #161, opened 2019-05-07
duration is in microseconds not milliseconds

yakir12:
The correct duration of the "annie_oakley.ogg" movie is 24224200 **microsecond** (24.2242 seconds), not milliseconds. Otherwise it would have been 6 hours and 43 minutes and 44.2 seconds long..!


ianshmean:
Good catch! Thanks

PR #164, opened 2019-05-09
Add helper error messages if Makie not loaded before

ianshmean:
The plan is to remove this functionality in favor of external view functions in `Makie`, `ImageView` etc.

but for the moment, this should help usage

### DIFF
--- a/src/VideoIO.jl
+++ b/src/VideoIO.jl
@@ -59,6 +59,21 @@ if Sys.isapple()
     end
 end
 
+#Helper functions to explain about Makie load order requirement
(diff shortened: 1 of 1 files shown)


timholy:
In case this comment got missed, I don't think you need to use a viewer package at all if you export videos as a [StreamingContainer](https://github.com/JuliaImages/ImageAxes.jl/blob/d558af3f7d728eff96943891aa9f0d4fb4793991/src/ImageAxes.jl#L189-L380).

ianshmean:
Indeed! I like that plan. Should’ve mentioned it up top.

PR #166, opened 2019-05-09
Initialize camera settings for every viewcam() to allow repeat usage

ianshmean:


src/VideoIO.jl
```diff
@@ -84,24 +100,8 @@ function __init__()
...
-        if Sys.iswindows()
-            DEFAULT_CAMERA_DEVICE[] = string(
-                "video=",
-                isempty(CAMERA_DEVICES) ? "0" : CAMERA_DEVICES[1]
-            )
-        end
-        if Sys.islinux()
-            DEFAULT_CAMERA_DEVICE[] = isempty(CAMERA_DEVICES) ? "" : CAMERA_DEVICES[1]
-        end
-
```
kmsquire: Should still call `init_camera_settings()` here, as a user can call `opencamera` without calling `viewcam()` (see #167).

PR #170, opened 2019-05-10
Tidying post BinaryProvider switch

ianshmean:
There are some redefinitions occurring due to back-support of prior ffmpeg versions.

Given we're now fixed to ffmpeg 4.1, shall we remove all the older version support?
If not we can keep this as a simple PR to remove that redefinition


ianshmean:
I've made an extreme tidy up version over at #171 
If that's considered too extreme, I can modify this one to be the right balance

PR #171, opened 2019-05-18
Tidying post BinaryProvider switch (extreme version)

ianshmean:
A more extreme version of https://github.com/JuliaIO/VideoIO.jl/pull/170

Based on the idea that, if needed for future development reference, all the old ffmpeg/libav wrappers can be accessed via git history (as explained in the new README, where v0.5.6 is referenced as it is the last released version with the BinDeps build), and don't need to be in the active package.

All mentions of libav in docs have been removed

### Package lines of code
**Pre-build**
master = 75,977  
tidy-extreme = 12,756 (-83%, i.e. ~6x smaller)

**Post-build** (MacOS)
master = 284,910  
tidy-exteme = 221,685 (-22%)


kmsquire:
There's certainly more that can be cleaned up with this change, but this is the biggest step.  LGTM!

PR #175, opened 2019-05-29
Correct variable name

CiaranOMara:


ianshmean:
Thanks for catching that!

PR #179, opened 2019-06-04
Fix bug with encoding messing up frame order

IanButterworth:
I can't reproduce the error locally, but @logankilpatrick was seeing an error with encoding a stack of ordered images.

@logankilpatrick It would be good if you could test that the code below produces a white spot rastering up and down the screen predictably on VideoIO#master and VideoIO#fix_encode_frameorder

```
using VideoIO

imgstack = []
for i=1:100
    img = UInt8.(ones(UInt8,10,10).*255) # Mostly white pixels
    img[i] = 0 # One black pixel
    push!(imgstack,img)
end

props = [:priv_data => ("crf"=>"22","preset"=>"medium")]

encodevideo("test.mp4", imgstack, AVCodecContextProperties = props)
```


logankilpatrick:
There's no white spot on VideoIO#master and VideoIO#fix_encode_frameorder

ianshmean:
The test video above plays in VLC but not quicktime. I think that Quicktime doesn't have proper support for single channel Gray8 h264 videos, which I hadn't noticed because I always use VLC.

This example creates an RGB example, which plays in quicktime.. but oddly the pixel jitters back and forth a few frames (in QT, not VLC):
```julia
using VideoIO, ColorTypes, FixedPointNumbers
imgstack = []
for i=1:2:(50*50)
    img = fill(RGB{N0f8}(1,1,1),(50,50))) # White
    img[i] = RGB{N0f8}(0,0,0) # One black pixel
    push!(imgstack,img)
end
props = [:priv_data => ("crf"=>"22","preset"=>"medium")]
encodevideo("test_rgb.mp4", imgstack, framerate=60, AVCodecContextProperties = props)
```
Even weirder is that this modified example of the above, with a little noise added, plays back properly in QT (and in VLC)..
```julia
using VideoIO, ColorTypes, FixedPointNumbers
imgstack = []
for i=1:2:(50*50)
    img = fill(RGB{N0f8}(1,1,1),(50,50)) .- (N0f8(0.3)*rand(RGB{N0f8},50,50)) # Mostly white pixels with noise
    img[i] = RGB{N0f8}(0,0,0) # One black pixel
    push!(imgstack,img)
end
props = [:priv_data => ("crf"=>"22","preset"=>"medium")]
encodevideo("test_rgb_noise.mp4", imgstack, framerate=60, AVCodecContextProperties = props)
```

🤷‍♂️ I think this is less of a videoio / ffmpeg issue, and more of getting into the weeds of _why is quicktime playback jittery_

@logankilpatrick I'm intrigued to see if you can find a MWE that plays back out of order in both QT and VLC, and if so, test it on this branch and master

logankilpatrick:
With "crf" set to 0 or 22?

logankilpatrick:
I will spend some time this weekend throwing together an example and testing!

ianshmean:
@logankilpatrick Did you happen to delve any further into this? There are a lot more tests now in `master` for frame order, pixel accuracy etc. so I'm intrigued as to whether you see the same issue still (and whether it's specific to QuickTime vs. VLC)

stillyslalom:
I'm seeing this on VideoIO#master. Videos play reliably on VLC, but on QuickTime, the first ~three frames are played forwards, the next ~two in reverse, etc. The number of frames between reversals isn't constant, and the reversals happen even when I step through the video frame by frame.

logankilpatrick:
Sorry for not getting a jump on this sooner. It’s in my to do list. Looks like it wasn’t fixed yet but I will dive in soon.

ianshmean:
What kind of frame rate are you both using? It sounds like you’re both using low ones. 

I’m wondering whether it’s that we’re stressing quicktime’s playback, rather than the actual encoding.

stillyslalom:
I've tried with between 10-20 frames per second, and if the problem's still present when I step through frame by frame, it seems likelier to be an encoding issue. I've previously been able to make error-less videos by piping frames to a system ffmpeg (although I haven't tried that with my current image stack).

ianshmean:
But if it’s an encoding issue, I don’t understand why it’s not a problem in VLC. Perhaps VLC has some insider knowledge on h264 that is using some red flag to trigger ignoring an erroneous frame order index. 

If I was to guess, I’d say the issue is occurring in the MUX step, but it’s hard to check because raw h264 stream files (the .stream file that’s temporarily written) can’t be opened in QuickTime. 

Obviously it needs fixing, I just don’t have any ideas how to..

stillyslalom:
This issue seems to rhyme: https://trac.ffmpeg.org/ticket/502

IanButterworth:
Should be fixed now, especially with #279 being merged

PR #180, opened 2019-06-05
Fix freebsd

ianshmean:


ianshmean:
FreeBSD is fixed! Nightly fails on freebsd, but only temporarily: https://github.com/JuliaLang/julia/issues/31559

PR #181, opened 2019-06-05
Precision testing for issues on ARM

ianshmean:
Frame comparison tests on ARM are known failures. This is to try to root out the source of the error


ianshmean:
New tests added to try and figure out why the reference image comparison is failing on ARM..

- UInt8 accuracy during read & encode: 
Tests that the truth video is read in accurately, where each pixel in a 16x16 frame is a unique UInt8 value
- Correct frame order when reading & encoding: 
Test that reading another truth video, where of 256 frames has pixels with the right incremental UInt8 value (the video frames fade black to white with time)

I also swapped out the png's for bmp's and it didn't fix the issue.

Not sure why it's happening given the above tests pass

ianshmean:
@kmsquire Any idea why this might be happening? I've tried testing for pixel value inaccuracy, out of order frames, and tried switching the bmp images.
 
I guess ImageMagick could still be messing up the reading of the png's on ARM..

ianshmean:
I've added arm ci to ImageMagick to see if tests pass: https://github.com/JuliaIO/ImageMagick.jl/pull/157

kmsquire:
@ianshmean, I think you're doing all of the right things. 

One idea in the same vein would be to save the image file itself in a different format (e.g., JLD), and see if the image being read in by ImageMagick (and/or ffmpeg) match.

ianshmean:
I moved over to JLD and the issue persisted unfortunately.. I'm out of ideas... I'm going to reset the comparisons to `@test_skip` for arm

Using JLD slows down the build process because it's BinDeps-based, so I've moved back to ImageMagick.

kmsquire:
Thanks for trying!

PR #186, opened 2019-06-12
add pixelaspectratio to play()

ianshmean:


ianshmean:
Thanks @yakir12. In testing this on MacOS we discovered a bug with viewcam() which slowed things down but is now fixed, and this is looking good

PR #187, opened 2019-06-15
type error due to aspect ratio, help function for the aspect

yakir12:
there was an error in playvideo due to a type mismatch in the resolution tuple. I then also fixed it so it will work with different aspect ratios by adding a convenience function for retrieving the stored aspect ratio. The works nicely for me but check to see it's good for you too.


ianshmean:
Great! Thanks

Issue #184, opened 2019-06-12
loglevel


yakir12:
It would be nice to set the `loglevel` to, say, `fatal` to avoid overzealous warnings about:
```julia
using VideoIO
file = download("https://vision-group-temporary.s3.amazonaws.com//a.MTS")
f = openvideo(file)
seek(f, 30.0)
```
spits out this:
```julia
[h264 @ 0x417e940] reference picture missing during reorder
[h264 @ 0x417e940] Missing reference picture, default is 2147483647
```

ianshmean:
I agree that we should suppress messages like this. There's one that pops up during `] test VideoIO` that's annoying.
`-loglevel fatal` is a `ffmpeg` executable-level setting. I can't immediately find one for the lower level functions. If you find one, please go ahead and submit a PR!

yakir12:
I'm blindly browsing in the ffmpeg git repo and trying to find clues. [This](https://git.ffmpeg.org/gitweb/ffmpeg.git/commitdiff/cdf17cf92b0f9c02951b7ea25d6a150000282bdf) seems relevant. Apparently we should be able to set this `FFREPORT` to 8 for showing only fatal errors (or `-8` for being totally silent). I don't know where or how but this is what it would look like:  
```
FFREPORT=level=8
```

yakir12:
We basically need to set the **environment variable** `FFREPORT` to one of the following nummerical values:
‘quiet, -8’

    Show nothing at all; be silent. 
‘panic, 0’

    Only show fatal errors which could lead the process to crash, such as an assertion failure. This is not currently used for anything. 
‘fatal, 8’

    Only show fatal errors. These are errors after which the process absolutely cannot continue. 
‘error, 16’

    Show all errors, including ones which can be recovered from. 
‘warning, 24’

    Show all warnings and errors. Any message related to possibly incorrect or unexpected events will be shown. 
‘info, 32’

    Show informative messages during processing. This is in addition to warnings and errors. This is the default value. 
‘verbose, 40’

    Same as info, except more verbose. 
‘debug, 48’

    Show everything, including debugging information. 
‘trace, 56’

I have no idea how...

giordano:
`ENV["FFREPORT"] = "8"`

yakir12:
> ENV["FFREPORT"] = "8"

didn't work (for me).

yakir12:
Maybe it needs to be specified in another environment than `ENV`.

giordano:
[...]

PR #189, opened 2019-06-17
Helper methods for set/get ffmpeg log level, and info on default

ianshmean:
Addresses #184.

Default log level is set to `AVUtil.AV_LOG_FATAL`, and can be set with:
```julia
VideoIO.loglevel!(VideoIO.AVUtil.AV_LOG_VERBOSE)
```
Docstrings:
```julia
help?> VideoIO.loglevel
  loglevel() -> String

  Get FFMPEG log level as a variable name string.

help?> VideoIO.loglevel!
  loglevel!(loglevel::Integer)

  Set FFMPEG log level. Options are:

    •    VideoIO.AVUtil.AV_LOG_QUIET

    •    VideoIO.AVUtil.AV_LOG_PANIC

    •    VideoIO.AVUtil.AV_LOG_FATAL

    •    VideoIO.AVUtil.AV_LOG_ERROR

    •    VideoIO.AVUtil.AV_LOG_WARNING

    •    VideoIO.AVUtil.AV_LOG_INFO

    •    VideoIO.AVUtil.AV_LOG_VERBOSE

    •    VideoIO.AVUtil.AV_LOG_DEBUG

    •    VideoIO.AVUtil.AV_LOG_TRACE
```

I'm not sure that this is a good idea.. but I've added an info message at import.. Perhaps we don't need that..?
```julia
julia> using VideoIO
[ Info: VideoIO: Low-level FFMPEG reporting set to minimal (AV_LOG_FATAL). See `?VideoIO.loglevel!()` for options
```

The MWE from #184 no longer spits out warnings on this branch
```julia
using VideoIO
file = download("https://vision-group-temporary.s3.amazonaws.com//a.MTS")
f = openvideo(file)
seek(f, 30.0)
```

It also suppresses the non-test-breaking warnings during `] test VideoIO`! woohoo!!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "VideoIO"
 uuid = "d6d074c3-1acf-5d4c-9a43-ef38773959a2"
-version = "0.6.4"
+version = "0.6.5"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ or encoded iteratively in custom loops.
 
 ### Platform Nodes: 
 
-- ARM: There is a known issue on ARM that results in small precision differences when reading/writing video files. 
-As such, tests for frame comparison are currently skipped on ARM. Issues/PRs welcome for helping to get this fixed
+- ARM: There is a known issue on ARM that results in very small and rare precision differences when reading/writing some video files. As such, tests for frame comparison are currently skipped on ARM. Issues/PRs welcome for helping to get this fixed.
 
 
 Installation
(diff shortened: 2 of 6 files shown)


yakir12:
I think that "daily" users of `VideoIO` won't feel that such a warning is necessary. For those that want to debug, develop, or test things, then they'll go looking for `ffmpeg`-style log-levels and find it in the documentation (which would be good to add). 

There are two bigger-picture related things to keep in mind here (in my humble opinion):
1. `VideoIO` ≠ `ffmpeg`: so we don't need to replicate every aspect of `ffmpeg` (e.g. the `ffmpeg` banner every time a command is run)
2. There will be other default settings that we will silently set for the users, sane default settings, so keeping the verbosity minimal at `import` is wise (since you'll need to make this decision many more times in the future).

yakir12:
see #191

ianshmean:
I like your logic. 
+1 on reasons to have proper docs

PR #190, opened 2019-06-17
Add test for SAR

yakir12:
Note that we should really add a video to the test videos that has a valid SAR that is not equal to 1. Such as [this](https://vision-group-temporary.s3.amazonaws.com//a.MTS) one, but that one won't stay at that url for much longer).


ianshmean:
Thanks! I think as @kmsquire has pointed out before, the structure of the tests needs a bit of an overhaul, but I think the location of this test is ok for now

PR #195, opened 2019-07-08
Add `gettime()` to read approximate current position in seconds

ianshmean:
@yakir12 Does this give what you want?

```
    gettime(s::VideoReader;video_stream::Integer=1)

Return timestamp of current position in seconds.
```

i.e.
```julia
using VideoIO
f = VideoIO.testvideo("annie_oakley")
v = openvideo(f)
@show gettime(v)
seek(v,10.0)
@show gettime(v)
seek(v,15.0)
@show gettime(v)
```
gettime(v) = 0.0
gettime(v) = 9.943266666666668
gettime(v) = 14.948266666666669

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -482,9 +482,9 @@ end
 Return timestamp of current position in seconds.
 """
 function gettime(s::VideoReader;video_stream::Integer=1)
-    stream = s.avin.video_info[video_stream].stream
+    time_base = stream.time_base
     frameindex = s.aVideoFrame[video_stream].pkt_dts   #av_frame_get_best_effort_timestamp(s.aVideoFrame)
-    return frameindex * (stream.time_base.num/stream.time_base.den)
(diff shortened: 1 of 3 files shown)


yakir12:
I mean, it looks great, but I'm still trying to figure out how to fetch a PR on git...

yakir12:
Yea, that is perfect! This way I can read all the frames from some known starting time to some other stopping time. Great! Thanks a lot!!!

asinghvi17:
Is this working?

ianshmean:
Yeah. But there is an open question about whether this should return in the same units as the duration functions, which return microseconds. 

I’d think having both return seconds would make most sense, but it would be a breaking change.

yakir12:
Breaking good, not "Breaking Bad".

ianshmean:
I've updated the duration function to return a float value in seconds (I'd have used `Dates.Second()` but they have to be discrete seconds)

PR #196, opened 2019-07-08
merge with #195

yakir12:
But check first that the way I retrieve the stream is ok. looks a bit shitty.


ianshmean:
I think you confused an old branch for your new one here @yakir12 

Assuming that's what happened, I'll close this as the purpose is now redundant

PR #199, opened 2019-07-12
Speeding up load time

ianshmean:
Master:
```
julia> @time using VideoIO
  5.601775 seconds (14.84 M allocations: 769.344 MiB, 5.81% gc time)
julia> exit()
...
julia> @time using VideoIO
  5.632717 seconds (14.84 M allocations: 769.345 MiB, 5.84% gc time)
```

This PR:
```
julia> @time using VideoIO
  3.208445 seconds (8.84 M allocations: 442.302 MiB, 7.95% gc time)
julia> exit()
...
julia> @time using VideoIO
  3.186645 seconds (8.84 M allocations: 442.302 MiB, 8.00% gc time)
```

@SimonDanisch Any other easy ideas?


asinghvi17:
Could drop Requires on master, VideoPlayer.jl should be coming along fairly soon (in a week or so)

asinghvi17:
Also,  is N0F8 available through color types?  May be possible to drop the dependency on fixedpointnumbers and just rely on colortypes

ianshmean:
For `Requires` I'll wait for `VideoPlayer.jl` to be ready 👍🏻

And `ColorTypes` can provide `N0f8` 👍🏻

ianshmean:
Relying on `ColorTypes` for `N0f8` has actually slowed it down a bit 🤔
```
julia> @time using VideoIO
  3.728758 seconds (8.82 M allocations: 440.791 MiB, 7.64% gc time)
julia> exit()
...
julia> @time using VideoIO
  3.712868 seconds (8.82 M allocations: 440.793 MiB, 7.59% gc time)
```

PR #200, opened 2019-07-12
Moving to FFMPEG.jl dep

ianshmean:
Work in progress.. currently fails because in the current regime, the output of ffmpeg needs to be collected through:

https://github.com/JuliaIO/VideoIO.jl/blob/40f21afd29492cbee7101f558fd5ae3b31ef6db5/src/util.jl#L31-L37


ianshmean:
Fixed with https://github.com/JuliaIO/FFMPEG.jl/pull/4 awaiting merging and releasing

asinghvi17:
Should be GTM now?  With compat on FFMPEG >= 0.2.0 for safety

ianshmean:
I’ve got the compat commit ready to push when the version is released on general 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

asinghvi17:
@ianshmean it's now merged.

PR #201, opened 2019-07-18
hotfix/encoding

CiaranOMara:
This hotfix addresses some undefined functions that I encountered in the `appendencode!` function. I'm not sure what these _usings_ do to your load times, so I've blanked out the affected times. I made a guess as to the intent of the undefined `restrict` function.


ianshmean:
Ah! Thanks for catching this. I didn’t realize that the color encoding functions weren’t being tested. We should also add tests for them here.

Restrict is actually an image transformation, to halve the resolution of the Cb &Cr color planes 

https://juliaimages.org/latest/function_reference/#ImageTransformations.restrict

CiaranOMara:
I found the restrict function. In terms of testing, I'd be keen to have a go at refactoring `prepareencoder` and `appendencode!` to work with multiple-dispatch, then test discrete functions. I'd like to leave this as a hotfix.

ianshmean:
Thanks for this. #202 incoming!

PR #215, opened 2020-01-25
Update reading.md

jonathan-durbin:
cam.framerate may be what was intended. This is *super* minor, but resulted in more wasted time for me than I'd like to admit...


ianshmean:
Thanks!

PR #244, opened 2020-06-15
Add frame skip & frame count functions

ianshmean:
The objective here is to provide the fastest way to 
1) skip through a frame 
2) skip through multiple frames 
3) get the most accurate & fastest count of remaining available frames

@rdeits @yakir12 What do you think about these?

I'll add tests if they look good

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -633,16 +633,20 @@ function skipframes(s::VideoReader, n::Int; throwEOF=true)
 end
 
 """
-    countframes(s::VideoReader)
+    counttotalframes(s::VideoReader)
     
-Count the number of frames remaining in the video by skipping through each frame.
+Count the total number of frames in the video by seeking to start, skipping through 
+each frame, and seeking back to the start.
 """
-function countframes(s::VideoReader)
+function counttotalframes(s::VideoReader)
+    seekstart(s)
     n = 1
     while true
-        skipframe(s, throwEOF = false) && return n
+        skipframe(s, throwEOF = false) && break
         n += 1
     end
+    seekstart(s)
+    return n
 end
 
 function eof(avin::AVInput)
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -3,6 +3,7 @@
 import Base: read, read!, show, close, eof, isopen, seek, seekstart
 
 export read, read!, pump, openvideo, opencamera, playvideo, viewcam, play, gettime
+export skipframe, skipframes, counttotalframes
 
 mutable struct StreamInfo
     stream_index0::Int             # zero-based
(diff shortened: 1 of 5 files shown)


src/avio.jl
```diff
@@ -601,6 +601,49 @@ end
...
+        skipframe(s, throwEOF=throwEOF) && return n
+    end
+end
+
+"""
+    countframes(s::VideoReader)
+    
+Count the number of frames remaining in the video by skipping through each frame.
+"""
+function countframes(s::VideoReader)
```
rdeits, line 640: Won't this result in the video being advanced to the end? That is, if I do `countframes(video)` and then try to `read(video)`, won't I get an EOF error?

ianshmean: Indeed. The user would want to `seekstart` afterwards. Or maybe that could be part of the function

ianshmean: I've changed this to `counttotalframes` which seeks to the start, counts, then seeks back to the start

ianshmean:
`counttotalframes` is currently under-counting by 3 (tested for n=5 and n=100)

```
using VideoIO
n = 100
imgstack = map(x->rand(UInt8,100,100),1:n)
props = [:priv_data => ("crf"=>"22","preset"=>"medium")]
encodedvideopath = VideoIO.encodevideo("testvideo.mp4",imgstack,framerate=30,AVCodecContextProperties=props, silent=true)

f = VideoIO.openvideo(encodedvideopath)
@show counttotalframes(f)
counttotalframes(f) = 97
```

yakir12:
Did I have a sea of similar problems with uncounted frames before... oy...

ianshmean:
It's weird because in the precision tests, we're producing a frame stack, encoding it to file, decoding it and checking every frame against the source framestack.. and those tests pass.

I could understand if it was n-1, perhaps due to the way that EOF is reported, but n-3 seems odd

ianshmean:
Ok, this is related to crf & keyframes. It works for lossless (but it's overcounting by one. I need to fix the EOF  handling):

```
using VideoIO

function testcount(n, crf)
        imgstack = map(x->rand(UInt8,100,100),1:n)
        props = [:priv_data => ("crf"=>crf,"preset"=>"medium")]
        encodedvideopath = VideoIO.encodevideo("testvideo.mp4",imgstack,framerate=30,AVCodecContextProperties=props, silent=true)

        f = VideoIO.openvideo(encodedvideopath)
        return counttotalframes(f)
end

testcount(5, "22")   # 2
testcount(5, "0")    # 6
testcount(5, "1")    # 2
```

ianshmean:
So it seems if `crf > 0` during encoding the total number of frames available to read back in is `n-4`.

Seems like an upstream issue?

ianshmean:
I'd be happy to merge this, if all agree

yakir12:
Sorry for the lack of input, looks good to me!

rdeits:
:+1:

PR #245, opened 2020-06-17
Try to introduce align to fix issue #243

### LINKED ISSUE #243: Strange "noise points"?
I found that `read` generates some "noise points" in the right region. More weirdly, it gives different results on Windows and Linux.

ffmpeg result (reference, both on Windows and Linux):

image

VideoIO on linux:
image

VideoIO on Windows:
image

Test video:
https://drive.google.com/file/d/1ThCzT_1PMkspxAktCY1bkvFFsG_IsNHc/view?usp=sharing

ffmpeg command:

```bash
ffmpeg -i out.mp4 frame%04d.png
```

VideoIO script:

```julia
using VideoIO
using FileIO

function fmt(x)
    s = string(x)
    "0"^(4-length(s)) * s
end

function split_frame(fname)
    io = VideoIO.open(fname)

    f = VideoIO.openvideo(io)

    img = read(f)

    let i=1
        while !eof(f)
            save("frame" * fmt(i) * ".png", img)
            i += 1
            read!(f, img)
        end
        save("frame" * fmt(i) * ".png", img)
    end
end

split_frame("out.mp4")
```

Do I need some extra options to avoid this? Or is there a "standard" format to make VideoIO work without such problems (The test video self is generated by ffmpeg with default parameter.)?

yiyuezhuo:
Fix #243 

Maybe performance will be damaged.


src/avio.jl
```diff
@@ -295,7 +295,14 @@ function VideoReader(avin::AVInput, video_stream=1;
                                  width, height, target_format,
                                  transcode_interpolation, C_NULL, C_NULL, C_NULL)
 
-    avpicture_fill(pointer(aTargetVideoFrame), target_buf, target_format, width, height)
+    align = 16
+    offset = 64 # length of data
```
ianshmean, line 299: Are these values correct across all resolutions/formats etc? Tests seemed to be passing after your first commit, so I guess that's decent evidence..

yiyuezhuo:
Performance comparsion using [given](https://drive.google.com/file/d/1ThCzT_1PMkspxAktCY1bkvFFsG_IsNHc/view?usp=sharing) video:

Original (wrong):

```
BenchmarkTools.Trial: 
  memory estimate:  243.42 KiB
  allocs estimate:  1326
  --------------
  minimum time:     21.015 ms (0.00% GC)
  median time:      21.151 ms (0.00% GC)
  mean time:        21.196 ms (0.06% GC)
  maximum time:     25.371 ms (0.00% GC)
  --------------
  samples:          236
  evals/sample:     1
```

fix_align:

```
BenchmarkTools.Trial: 
  memory estimate:  242.86 KiB
  allocs estimate:  1320
  --------------
  minimum time:     22.293 ms (0.00% GC)
  median time:      22.447 ms (0.00% GC)
  mean time:        22.489 ms (0.06% GC)
  maximum time:     25.591 ms (11.94% GC)
  --------------
  samples:          223
  evals/sample:     1
```

src/avio.jl
yiyuezhuo: offset = 64 length of [AVFrame.data](https://ffmpeg.org/doxygen/2.7/structAVFrame.html), 8 byte = pointer length in 64 bit system and AV_NUM_DATA_POINTERS=8, thus 64. Maybe I should add 32 bit compatibility. `align=16` is value used by [PyAV](https://github.com/PyAV-Org/PyAV/commit/673f072e6a8552b6c5c8697d9e2d47b65ea9dc34#diff-dc208ab0752ca9c4f2bf1fc7c7fc3efc), they tried custom specifying, forced 8 and finally use forced 16.

ianshmean: Thanks for the detail. I think adding in-line comments to explain & a link to the PyAV commit would be helpful, otherwise they're a bit obscure

ianshmean: And I just noticed that the 32-bit windows CI is failing. We should make the offset dependent on word size

yiyuezhuo: I added `WORD_SIZE` but I don't have 32-bit system to run test locally. Wait for CI result.

ianshmean:
Sorry for the delay on this. I'm hoping to get at least one more set of eyes on it, given it's core to the read functionality. If I don't by the end of this week, I'll merge and release

PR #249, opened 2020-08-01
Allow encoding with rational frame rates

galenlynch:
While ffmpeg allows the use of rational frame rates, VideoIO.jl assumes that frame rates can only be integers. Many video capture devices do not have exactly integer frame rates, making it difficult to process their output with VideoIO.jl. I have therefore made simple modifications to the encoding code to allow rational frame rates.


ianshmean:
Looks great, thanks for the addition and for the tests

PR #250, opened 2020-08-01
Increase efficiency of accessing members of struct pointers

galenlynch:
The current code to access a member of a struct pointer uses a package function, `fieldposition`, that has identical functionality to a function provided in base, `fieldindex`. While this latter function is not exported, it is documented and will hopefully remain a part of Julia's API. I imagine that the base function is more performant than the implementation in `fieldposition`, which compares each field name to the target field name in order to find the position of a field. As such, I have replaced calls to `fieldposition` with calls to `Base.fieldindex`, and deleted the function `fieldposition`.

Additionally, when setting a member of a struct pointer, the current implementation calculates the pointer position, wraps it in a julia array with `unsafe_wrap`, and then sets the first element of that array to the desired value. The same result is achieved, perhaps more simply, by using `unsafe_store!` and not using an intermediate array. I have therefore replaced the call to `unsafe_wrap` and subsequent code with `unsafe_store!`.

While I do not know if the existing code causes a measurable decrease in performance, using the base functions reduces the amount of code, and most likely makes it faster.

### DIFF
--- a/src/ffmpeg/AVUtil/v56/LIBAVUTIL.jl
+++ b/src/ffmpeg/AVUtil/v56/LIBAVUTIL.jl
@@ -54,10 +54,10 @@ function AVFrame()
   ts = AVFrame.types
   parms = [T <: Ptr ? C_NULL : zero(T) for T in ts]
 
-  fmt_pos = fieldposition(AVFrame, :format)
+  fmt_pos = Base.fieldindex(AVFrame, :format)
   parms[fmt_pos] = -one(ts[fmt_pos])
 
-  pts_pos = fieldposition(AVFrame, :pts)
+  pts_pos = Base.fieldindex(AVFrame, :pts)
   parms[pts_pos] = AV_NOPTS_VALUE
 
   AVFrame(parms...)
--- a/src/util.jl
+++ b/src/util.jl
@@ -1,27 +1,18 @@
 # Helpful utility functions
 
-function fieldposition(::Type{T}, name::Symbol) where T
-    field_pos = findfirst(isequal(name), fieldnames(T))
-    if field_pos === nothing
-        throw(ErrorException("The type `$T` does not contain field $name!"))
-    end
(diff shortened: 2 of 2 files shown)


ianshmean:
Good catch. It would be good to benchmark this before merging, I want to be sure that we don't impede performance (I agree with your logic, but still worth checking).
Would you mind constructing `BenchmarTools.@btime` tests for frame read/write and compare to master? thanks!

galenlynch:
I only measured the write functionality, which is the majority of this PR (though reads are also affected), but the PR is about 10 times faster, and uses less memory:

on master:
```julia
julia> using VideoIO, BenchmarkTools

julia> codec_name = "libx264";

julia> codec = avcodec_find_encoder_by_name(codec_name);

julia> apCodecContext = Ptr{AVCodecContext}[avcodec_alloc_context3(codec)];

julia> width = 1024;

julia> @benchmark av_setfield(apCodecContext[1],:width,width)
BenchmarkTools.Trial: 
  memory estimate:  1.77 KiB
  allocs estimate:  8
  --------------
  minimum time:     3.070 μs (0.00% GC)
  median time:      3.141 μs (0.00% GC)
  mean time:        3.364 μs (1.17% GC)
  maximum time:     146.271 μs (95.39% GC)
  --------------
  samples:          10000
  evals/sample:     8
```

for this PR:

```julia
...
julia> @benchmark av_setfield(apCodecContext[1],:width,width)
BenchmarkTools.Trial: 
  memory estimate:  96 bytes
  allocs estimate:  6
  --------------
  minimum time:     384.736 ns (0.00% GC)
  median time:      401.701 ns (0.00% GC)
  mean time:        414.897 ns (0.39% GC)
  maximum time:     4.677 μs (89.38% GC)
  --------------
  samples:          10000
  evals/sample:     201

```

ianshmean:
Ok great. Thanks

PR #254, opened 2020-08-04
Allow read! to work with regular arrays, not just PermutedDimsArrays

### LINKED ISSUE #253: Undocumented change to `read!` between VideoIO versions 0.7.1 and 0.7.2 causes errors
I have been using 0.7.1 for a while, but noticed that my client code is broken on master. After a bit of git-blaming, I found that [this line in commit eaf4dbc](https://github.com/JuliaIO/VideoIO.jl/blob/eaf4dbc41f0fa4880281c06060e8fc9d94460cd9/src/avio.jl#L438) (part of PR #245) is causing breakage for me, although I may have been abusing the API and it's just now come back to haunt me. 

Specifically, I mostly work with images and frames of movies in a "scan-line major", or image-x major, memory format as many libraries expect that, including FFMPEG itself. However, VideoIO instead mostly assumes a row major format, where rows of a matrix are equated with image y dimension. Until eaf4dc, I simply used the parent array underlying the `PermutedDimsArray` view returned by `VideoIO.read(r::VideoReader)`, and passed that parent array to subsequent calls to `VdeoIO.read!(r::VideoReader, arr::VidArray)` to read a video. 

While the documentation says nothing about what _type_ of array `VideoIO.read!` expects, the fact that `VidArray{T,N} = Union{Array{T,N},PermutedArray{T,N}}` led me to believe that passing it regular arrays was fine as far as the supported API was concerned, and obviously `read!` did not care since no permutation was being done, and in the end all the `read!` function was doing with `buf` was passing a pointer to the image-x major buffer to FFMPEG.

eaf4dc directly accesses the `parent` field of the `buf::VidArray` parameter, which is fine if `buf` is a `PermutedArray` but causes an error if it is the other portion of the `VidArray` union, namely a normal array. This line could be replaced by `parent(buf)`, which would work in either case, or could also just be eliminated since `pointer(buf) === pointer(parent(buf))`, and nothing else is done with the parent array. I'll make a PR that does the latter soon, but that does not address the potential breakage that users will experience if they check out VideoIO 0.7.2. I'm unsure how to fix that breakage, so I'm positing this issue independent of my upcoming PR that will fix the underlying bug. Also some clarification of what is considered supported behavior would probably help: is `read!` with normal arrays supported?

galenlynch:
Commit eaf4dbc fixed a bug where FFMPEG's line size alignment requirements were not met, but in doing so unintentionally restricted the types of Julia arrays that would would work with VideoIO's `read!` from the original `VidArray = Union{PermutedArray,Array}` to just `PermutedArrays`. While the default return type of VideoIO's `read` is a `PermutedDimsArray`, this commit caused regular arrays to throw an error when used with `read!`, and in effect changed the (undocumented) API of `read!` between VideoIO versions v0.7.1 and v0.7.2. Furthermore, the fixed `read!` that respects FFMPEG's SIMD alignment requirements should work with either a normal `Array`, or instead a `PermutedDimsArrays` view of that array; the memory layout of both types of buffers are identical.

I have removed the line that caused the error when used with normal arrays: an attempt to access the `parent` field of the buffer. I have also changed the source of frame width and height information from the size of the permuted buffer to the width and height information given by the transcoder. This switch eliminates another bug when copying lines from the AVFrame managed by ffmpeg into the buffer used by Julia client code when that second buffer is a regular array. In addition to fixing this bug, the FFMPEG transcoder is a more authoritative source of information about frame dimensions, and indeed no checking was done to ensure that the height and width of the buffer matches the height and width of the transcoder.

I have also eliminated the temporary variables `o_stride` and `o_size`, which are never used.

I have added a bit of documentation to VideoIO's `read!` function, describing the valid types of buffers for `read!`.

Finally, I have added tests to ensure `read!` works with both `PermutedDimsArray`s as well as `Array`s. In the process, I have refactored some code that would otherwise be repeated for these tests into helper functions.

Closes #253

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -433,9 +433,6 @@ function retrieve!(r::VideoReader{TRANSCODE}, buf::VidArray{T}) where T <: Eight
 
     i_buf = Ptr{UInt8}(pointer(r.transcodeContext.aTargetVideoFrame))
     i_stride = unsafe_load(pointer(r.transcodeContext.aTargetVideoFrame)).linesize[1]
-    o_stride = unsafe_load(pointer(r.transcodeContext.aTargetVideoFrame)).width * bytes_per_pixel
-    o_size = unsafe_load(pointer(r.transcodeContext.aTargetVideoFrame)).height * o_stride
-    o_buf = buf.parent
 
     _p = r.transcodeContext.aTargetVideoFrame[1].data[1]
     if t.target_bits_per_pixel == 8
@@ -444,12 +441,13 @@ function retrieve!(r::VideoReader{TRANSCODE}, buf::VidArray{T}) where T <: Eight
         p = Ptr{RGB{N0f8}}(Int(_p))
     end
 
-    height, width = size(buf)
+    height = t.height
+    width = t.width
 
-    o_buf_p = pointer(o_buf)
-    
+    buf_p = pointer(buf)
+
     for h in 1:height
-        op = o_buf_p + (h-1)*width*bytes_per_pixel
+        op = buf_p + (h-1)*width*bytes_per_pixel
         ip = p + (h-1) * i_stride
         unsafe_copyto!(op, ip, width)
     end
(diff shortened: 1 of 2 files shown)


ianshmean:
Thanks for this. We should get a release out so I'll merge and tag

PR #257, opened 2020-08-11
Fix one instance of type instability within VideoReader

galenlynch:
In `VideoReader`'s constructor, `aspect_ratio` is either set to the aspect ratio from the codec context, which is converted into a `Rational{Cint}`, or instead `0 // 1` if the denominator is zero, which is a `Rational{Int}`. This causes unceccessary type instability, which can be fixed by instead making `aspect_ratio = Cint(0) // Cint(1)` if the denominator is zero.


ianshmean:
CI failure here was a download blip

PR #260, opened 2020-08-17
Add utility function `get_number_frames`

galenlynch:
There is currently no way to query a video container to determine the number of video frames that it contains, besides decoding the video and counting the frames. While this is the best you can do for some container formats, other container formats can report the number of frames without decoding each video frame. I have added a new function, `get_number_frames`, which will query the container and return the number of frames, if applicable, or return `nothing` otherwise. This function will not decode the video stream in order to determine the number of frames.

I have additionally added simple tests of this function, a documentation string, and added it to the docs.

### DIFF
--- a/docs/src/reading.md
+++ b/docs/src/reading.md
@@ -147,3 +147,6 @@ VideoIO.get_time_duration
 ```@docs
 VideoIO.get_duration
 ```
+```@docs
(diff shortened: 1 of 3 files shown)


ianshmean:
Nice 👍

PR #261, opened 2020-08-17
Add iterator interface to VideoReader objects

galenlynch:
I have added an iterator interface to the `VideoReader` type, making collecting all frames of a video as simple as:

```julia
f = openvideo(fname)
imgs = collect(f)
```

The iterator returns a new array for each frame, as I think mutating the same array on subsequent calls to `iterate` would be surprising and lead to strange behavior when calling functions such as `collect`. Note that the VideoReader iterator is mutable, like Channel or Stateful iterators in Julia, and will resume iteration where it was last stopped, instead of at the beginning. I have added some text in the documentation explaining this behavior, and also added some tests.

### DIFF
--- a/test/avio.jl
+++ b/test/avio.jl
@@ -162,11 +162,7 @@ end
 
             ## Test that iterator is mutable, and continues where iteration last
             ## stopped.
-            i = 0
-            for _ in v
-                i += 1
-            end
-            @test i == 0
+            @test iterate(v) === nothing
         end
     end
 
--- a/test/avio.jl
+++ b/test/avio.jl
@@ -123,12 +123,13 @@ end
     for name in VideoIO.TestVideos.names()
         # TODO: fix me?
         (startswith(name, "ladybird") || startswith(name, "NPS")) && continue
-        @testset "Testing $name" begin
-            first_frame_file = joinpath(testdir, swapext(name, ".png"))
-            first_frame = load(first_frame_file)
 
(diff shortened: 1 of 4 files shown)


galenlynch:
I'm confused by these testing errors.

test/avio.jl
```diff
@@ -147,13 +148,23 @@ end
...
             end
+
+            # Iterator interface
+            VT = typeof(v)
+            @test Base.IteratorSize(VT) === Base.SizeUnknown()
+            @test Base.IteratorEltype(VT) === Base.EltypeUnknown()
+
+            VideoIO.seekstart(v)
+            @test length(collect(v)) == VideoIO.TestVideos.videofiles[name].numframes
+            @test length(collect(v)) == 0
```
ianshmean, line 159:
```suggestion
            imgstack = collect(v)
            @test length(imgstack) == VideoIO.TestVideos.videofiles[name].numframes
            imgstack = nothing
            @test length(collect(v)) == 0
            GC.gc()
```

galenlynch: Yeah good points... if CI is memory limited I probably should be using `collect` at all. I'll do some other test of the iterator interface that doesn't require storing all of the frames in memory at the same time.

ianshmean:
Looks good now. Shall we merge?

galenlynch:
Fine by me!

PR #268, opened 2020-10-19
Remove `permuteddimsview`, which is deprecated

galenlynch:
`permuteddimsview` has been deprecated in ImageCore.jl, in favor of `PermutedDimsArray` in Base. I have replaced `permuteddimsview` with `PermutedDimsArray` in this package to avoid deprecation warnings.

### DIFF
--- a/src/VideoIO.jl
+++ b/src/VideoIO.jl
@@ -1,10 +1,10 @@
 module VideoIO
 
-using Libdl                                             #0.227492 seconds
-using Requires, Dates, ProgressMeter                    #0.001749 seconds
-using ImageCore: channelview, rawview #0.723065 seconds
-using ColorTypes: RGB, Gray, N0f8, YCbCr                #0.529263 seconds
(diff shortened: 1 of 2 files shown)


src/VideoIO.jl
```diff
@@ -2,7 +2,7 @@ module VideoIO
 
 using Libdl                                             #0.227492 seconds
 using Requires, Dates, ProgressMeter                    #0.001749 seconds
-using ImageCore: permuteddimsview, channelview, rawview #0.723065 seconds
+using ImageCore: channelview, rawview #0.723065 seconds
 using ColorTypes: RGB, Gray, N0f8, YCbCr                #0.529263 seconds
 using ImageTransformations: restrict                    #3.156594 seconds!!
```
ianshmean, line 7: While we're here, would you mind removing those timing comments. They're likely well out of date

galenlynch: Sure

ianshmean: Thanks!

PR #277, opened 2020-11-19
Add code for non-transcoded video streams

### LINKED ISSUE #111: retrieve! for VideoReader{NO_TRANSCODE} does nothing
I am trying to get a NO_TRANCODE image, however read / retrieve returns a Void.
Reading the code I found weird that retrieve! does not return the buf.

```julia
function retrieve!{T<:EightBitTypes}(r::VideoReader{NO_TRANSCODE}, buf::VidArray{T})
    while !have_frame(r)
        idx = pump(r.avin)
        idx == r.stream_index0 && break
        idx == -1 && throw(EOFError())
    end

    if !bufsize_check(r, buf)
        error("Buffer is the wrong size")
    end
end
```

I believe that this function does nothing on practice, as there is no copy of the image buffer to buf; neither this buffer is returned.

dawbarton:
Fixes #111 and enables frames to be grabbed from a webcam without transcoding. This is particularly useful when only luminance values are required and the transcoding to RGB is actually unhelpful. 

I use this within a Raspberry Pi computer vision system for control. As such, speed is important - without transcoding I can grab a frame in 1.5ms whereas with transcoding it takes 63.5ms.

The data is returned in a UInt8 array rather than a more specifically typed array but this seems to be in line with the original (unimplemented) intent. To interpret this data, the specific webcam pixel format needs to be known but that is relatively trivial (see `cam = VideoIO.opencamera(;transcode=false); cam.format` and compare against the `AVPixelFormat` values in `libavutil_h.jl`.

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -405,7 +405,7 @@ function retrieve(r::VideoReader{NO_TRANSCODE}) # false=don't transcode
 
     # TODO: set actual dimensions ?
     buf_sz = avpicture_get_size(r.format, r.width, r.height)
-    buf = Array(UInt8, buf_sz)
+    buf = Array{UInt8}(undef, buf_sz)
 
     retrieve!(r, buf)
 end
@@ -476,6 +476,10 @@ function retrieve!(r::VideoReader{NO_TRANSCODE}, buf::VidArray{T}) where T <: Ei
     if !bufsize_check(r, buf)
         error("Buffer is the wrong size")
     end
+
+    unsafe_copyto!(pointer(buf), r.aVideoFrame[1].data[1], sizeof(buf))
+
+    return buf
 end
 
 # Utility functions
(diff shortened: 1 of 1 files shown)


yakir12:
on my RPI this:
```julia
using VideoIO, BenchmarkTools
cam = VideoIO.opencamera()
img = read(cam)
@btime read!($cam, $img)
```
results in 11.039 ms.
Using this PR, and `transcode=false` results in 845.115 μs! A ~13 times speedup!

ianshmean:
For anyone doing a test like this ^ in the future, you may want to 
```
@btime read!($cam, $img) setup=(sleep(0.1))
```
to give the camera time to make a new frame available, if the camera buffer waits for new, otherwise the timing may depend on framerate

ianshmean:
Thanks @dawbarton and thanks for testing @yakir12

PR #286, opened 2020-12-02
Remove 'archived' gen files

galenlynch:
These files do not appear to be used, nor do they seem to have been used to generate the libav/ffmpeg bindings in VideoIO.

In case they really are just "archive" code, and not needed for something else, I am submitted this PR to remove them.


IanButterworth:
We can always retrieve these from the history anyway, so I think you're right to suggest removing them

PR #296, opened 2021-01-25
Fix CI in #279

ianshmean:


giordano:
Wrong target branch?

.drone.jsonnet
```diff
@@ -17,6 +17,6 @@ local Pipeline(os, arch, version) = {
 };
 
 [
-    Pipeline("linux", "arm",   "1.4"),
-    Pipeline("linux", "arm64", "1.4")
+    Pipeline("linux", "arm",   "1.5"),
```
giordano, line 20: There is no build for Julia v1.5 on armv7l.  The latest ones are 1.3.1 and v1.4.1

ianshmean:
>Wrong target branch? 

I did intend to PR into #279 but it doesn't appear as an option.. which might be because maintainers aren't allowed to modify it? or perhaps branches on forks never allowed it?
So this is just a testing PR now.

Thanks for the drone tip

.drone.jsonnet
giordano: To be clear, this file is not used by Drone directly, only `.drone.yml` matters, and that one didn't change.  This is useful to generate the YAML file automatically though, which is much more cumbersome to do when you want to support multiple Julia versions and/or architectures.  Note also v1.4.1 != v1.4, because there has been v1.4.2

giordano:
> I did intend to PR into #279 but it doesn't appear as an option.. which might be because maintainers aren't allowed to modify it? or perhaps branches on forks never allowed it?

You need to open a PR in https://github.com/galenlynch/VideoIO.jl, targeting the [`add_10bit_gray_encoding`](https://github.com/galenlynch/VideoIO.jl/tree/add_10bit_gray_encoding) branch

ianshmean:
You're right https://github.com/galenlynch/VideoIO.jl/pull/1

PR #300, opened 2021-03-14
Fix typedef, tidy up tests

IanButterworth:
Updates to CI setup, removing redundant files, reorganizing tests because we had a lot going on in a single file


IanButterworth:
@galenlynch It seems something about ReinterpretArray has changed in 1.6

```
julia> Base.ReinterpretArray{UInt8, 2, N6f10, Base.ReinterpretArray{N6f10, 2, Gray{N6f10}, Matrix{Gray{N6f10}}, true}, false} <: StridedArray
false
```

The problem seems to happen because the outer function
https://github.com/JuliaIO/VideoIO.jl/blob/0388303991844d602e90b22084ab0a969eb9b531/src/avframe_transfer.jl#L272-L273 
has a looser definition for img than the child
https://github.com/JuliaIO/VideoIO.jl/blob/0388303991844d602e90b22084ab0a969eb9b531/src/avframe_transfer.jl#L199-L201

IanButterworth:
Relaxing to the same `img::AbstractArray{UInt8}` typedef  for `transfer_img_bytes_to_frame_plane!` seems to fix it. I'm guessing the narrowing typedef here was just an unintentional inconsistency that slipped through given for some reason it's ok on 1.5

IanButterworth:
I'm going to merge because I want to move on to the next PR, but let me know if the typedef change wasn't appropriate @galenlynch

galenlynch:
I think relaxing the typebounds is fine, IRC the real safety check is the call to `stride`. I'll try to remember why I put that type bound there in the first place.

IanButterworth:
Investigating an upstream fix https://github.com/JuliaLang/julia/pull/40027

PR #304, opened 2021-03-15
Naming and settings fixes

IanButterworth:
@galenlynch if we rename `container_settings`->`container_options` do we also want to:

change
```
 container_settings::SettingsT = (;),
 container_private_settings::SettingsT = (;),
 encoder_settings::SettingsT = (;),
 encoder_private_settings::SettingsT = (;),
```

to

```
 container_options::OptionsT = (;),
 container_private_options::OptionsT = (;),
 encoder_options::OptionsT = (;),
 encoder_private_options::OptionsT = (;),
```

Note `OptionsT`

### DIFF
--- a/docs/src/writing.md
+++ b/docs/src/writing.md
@@ -9,8 +9,8 @@ Videos can be encoded directly from image stack using `VideoIO.save(filename::St
 The entire image stack can be encoded in a single step:
 ```julia
 import VideoIO
-encoder_settings = (crf=23, preset="medium")
-VideoIO.save("video.mp4", imgstack, framerate=30, encoder_settings=encoder_settings)
+encoder_options = (crf=23, preset="medium")
+VideoIO.save("video.mp4", imgstack, framerate=30, encoder_options=encoder_options)
 ```
 
 ```@docs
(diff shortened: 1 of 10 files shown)


galenlynch:
That seems right to me!

galenlynch:
One comment about the documentation: true lossless encoding no longer requires libx264rgb. I haven't tested that's it's completely lossless, but VideoIO now supports YUV444 encoding, not just YUV420, which at least avoids the chroma subsampling.

IanButterworth:
I kept the renames as individual commits, so let me know if any should be dropped

> true lossless encoding no longer requires libx264rgb

This would be good to write up for the docs

galenlynch:
Sure!

galenlynch (approved):
I made a PR against this branch with my suggestions.

galenlynch:
> One comment about the documentation: true lossless encoding no longer requires libx264rgb. I haven't tested that's it's completely lossless, but VideoIO now supports YUV444 encoding, not just YUV420, which at least avoids the chroma subsampling.

I just tested it and `libx264` is not lossless even when `crf=0`. I don't know this is due to some colorspace problem, or instead something else on our end, but there's lots of off-by-one errors. I could probably also try with actual mpeg-range input, but that's probably beyond the pale for top-level documentation.

IanButterworth:
Even with `color_range=2`? That's required for Gray, so might be needed for RGB

galenlynch:
Yeah. It goes through sws_scale for RGB->YUV which is notoriously complicated so it's perhaps not surprising that it's not lossless without spending the time to fine-tune the sws_scale parameters.

Issue #313, opened 2021-04-07
UndefVarError bits_per_pixel when transforming to gray during decoding


Octogonapus:
When opening a raw Gray video in Gray color space, the [Gray transform block of VideoReader](https://github.com/JuliaIO/VideoIO.jl/blob/3eac5aaf597405fc6861e5867cb41bd07c396162/src/avio.jl#L311) results in an UndefVarError:
```
ERROR: LoadError: UndefVarError: bits_per_pixel not defined
Stacktrace:
 [1] VideoIO.VideoReader(avin::VideoIO.AVInput{String}, video_stream::Int64; transcode::Bool, target_format::Int32, pix_fmt_loss_flags::Int64, target_colorspace_details::Nothing, allow_vio_gray_transform::Bool, swscale_options::NamedTuple{(), Tuple{}}, sws_color_options::NamedTuple{(), Tuple{}})
   @ VideoIO ~/.julia/packages/VideoIO/B6VOI/src/avio.jl:311
 [2] #VideoReader#19
   @ ~/.julia/packages/VideoIO/B6VOI/src/avio.jl:339 [inlined]
 [3] #openvideo#20
   @ ~/.julia/packages/VideoIO/B6VOI/src/avio.jl:588 [inlined]
```

The video is opened in this way:
```julia
VideoIO.openvideo(video_path, target_format = VideoIO.AV_PIX_FMT_GRAY8)
```

Video description from ffmpeg:
```
Input #0, avi, from 'video.avi':
  Metadata:
    encoder         : Lavf54.35.100
  Duration: 00:00:26.52, start: 0.000000, bitrate: 629158 kb/s
    Stream #0:0: Video: rawvideo (Y800 / 0x30303859), gray, 1280x1024, 629541 kb/s, 60 fps, 60 tbr, 60 tbn, 60 tbc
```

I am using VideoIO v0.9.1 and Julia v1.6.0.

PR #314, opened 2021-04-07
Fix typo for bit_depth

IanButterworth:
@galenlynch this appears to be a typo. Can you confirm?

Evidently our tests aren't covering this branch.

Fixes https://github.com/JuliaIO/VideoIO.jl/issues/313


galenlynch:
Whoops my bad. That is definitely a typo, but the correct value is `bit_depth`, not `padded_bits_per_pixel`. I'll make a PR against your branch.

IanButterworth:
@galenlynch this slipped through. Do you recall if it's still correct? I assume so, but confirmation would be appreciated

galenlynch:
Hey sorry I've been AWOL. Will check it today.

PR #315, opened 2021-04-07
add RGB -> monochrome read test

IanButterworth:
Tests the failure in #313 (this should fail on master)

Should be merged with the correct failure before #314


galenlynch:
Thanks for adding this much needed test. If you have the bandwidth, I think it would make sense to test for accuracy, as well as making sure it does not error.

IanButterworth:
Unfortunately this still doesn't hit the problematic code path. We need a test that hits 
https://github.com/JuliaIO/VideoIO.jl/blob/33200c28ad823759f7548ca34c5beaf3524b9761/src/avio.jl#L311

PR #331, opened 2021-08-12
Enable adjustable codec multithreading (significant read and write speed up!)

IanButterworth:
I had assumed that ffmpeg/h.264 automatically handled multithreading, but it seems not to be the case.
The small change in this PR is based on https://stackoverflow.com/questions/43251612/ffmpeg-how-to-use-multithreading

I propose that the default is set to `Threads.nthreads()` but that users can tune it as they wish.

## Master
```julia
julia> function foo()
    f = openvideo("vid.mp4")
    i = 1
    img = read(f)
    i += 1 
    while !eof(f)
        read!(f, img)
        i += 1
    end
    close(f)
    @show i
end


julia> @time foo();
i = 1789
 27.231079 seconds (1.91 k allocations: 9.032 MiB, 0.03% gc time)
```
## This PR
Where `Threads.nthreads() == 16`
```julia
julia> @time foo();
i = 1789
  3.339067 seconds (1.91 k allocations: 9.032 MiB)
```

Saves about 1 minute (6 -> 5 mins on ubuntu) in CI with only 2 threads

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,6 +26,8 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
       - uses: julia-actions/julia-runtest@master
+        env:
+          JULIA_NUM_THREADS: 2
       - uses: julia-actions/julia-processcoverage@master
       - uses: codecov/codecov-action@master
         with:
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -248,7 +248,8 @@ function VideoReader(avin::AVInput{I}, video_stream = 1;
                      target_colorspace_details = nothing,
                      allow_vio_gray_transform = true,
                      swscale_options::OptionsT = (;),
-                     sws_color_options::OptionsT = (;)) where I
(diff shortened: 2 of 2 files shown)


IanButterworth:
Turns out it also helps encoding, but to a lesser extent.. only halves the time.. :)

```julia
julia> imgstack = map(_->rand(UInt8, 1000,1000), 1:1000);

julia> @time VideoIO.save("vid.mp4", imgstack, thread_count = 1);
  3.999581 seconds (30.01 M allocations: 610.917 MiB, 2.31% gc time)

julia> @time VideoIO.save("vid.mp4", imgstack); # defaults to thread_count = Threads.nthreads() == 6
  2.025728 seconds (30.01 M allocations: 610.917 MiB, 4.35% gc time)
```

IanButterworth:
github appear to have broken this branch and renamed it to a new branch called `IB/decode_threading` with capital initials. I'm no longer able to push here so I just opened #332 to replace this

PR #333, opened 2021-08-26
Update Clang wrapper generator and wrapper code

### LINKED ISSUE #287: Cannot generate FFMPEG bindings
It seems that the script to generate the ffmpeg bindings no longer runs, at least on Julia 1.5.2. 

It seems to use the Clang.jl interface that was deprecated in their version v0.8. Unfortunately Clang.jl seems to have a bug in the v0.8 range that prevents it from loading on my machine. Trying to use earlier versions (v0.7) no longer works on my machine because it seems to use my system's clang library, which was compiled with a newer version of GLIBCXX than the libstdc++ that is shipped with Julia.

melonedo:
The generator uses the master branch of Clang.jl (add https://github.com/JuliaInterop/Clang.jl). Most troubles come from `avptr`, but hopefully I can add a compatibility layer for this.
Currently it is impossible to split different modules based on header structures because there might be problems about where to put type definitions, so I collected all of the wrappers at `libffmpeg.jl`.

May fix #287.
cc @gnimuc


melonedo:
Oops the tests failed on windows because I did not import `ffmpeg`, now that is fixed.
@IanButterworth
cc @gnimuc

Gnimuc:
It's not hard to manually filter out those duplicated decls, see https://github.com/maleadt/LLVM.jl/blob/790a3d4b231f1fd41848710a37dd190d44e45740/res/wrap_llvmextra.jl#L37.

Gnimuc:
Looks like GitHub CI for Windows is temporarily down.

IanButterworth:
@galenlynch I wanted to check if you have a view on this? IMO we should merge

IanButterworth:
@melonedo @Gnimuc One question.. in #340 I'm adding error codes to julia errors. Is there a way to automate converting those codes back to the ffmpeg code symbols, i.e. `ENOMEM`? 

This change doesn't expose those symbol values, AFAICT

melonedo:
These error codes are not all defined by FFMPEG. Instead, some of them are just "standard" error codes defined in `errorno.h`, e.g. `ENOMEM`. I suppose these constants are already exposed somewhere. Simple searching found an old version of julia had them: https://github.com/Wilfred/julia/blob/c258b4c3aca5c17e7c7069f28776fe62e0ec5e73/base/errno.jl. 
Some of them are defined in FFMPEG, these are already exported at https://github.com/JuliaIO/VideoIO.jl/blob/bd4061a0bb670b40875324bf0ac3cc1abc55ee52/lib/libffmpeg.jl#L21397-L21449.

IanButterworth:
Of course. I picked a bad example :)

That's great. Following this PR we can write an error code -> symbol converter for the error messages.

@galenlynch I'll merge tomorrow if I don't hear otherwise :)

PR #351, opened 2022-04-15
Lazy-load cameras to improve package load time

IanButterworth:
Defers populating available cameras and identifying the default until the user requests cameras, which helps package load time.

Before
```
julia> @time using VideoIO
  2.679612 seconds (3.20 M allocations: 237.017 MiB, 3.63% gc time, 34.81% compilation time)
```

This PR
```
julia> @time using VideoIO
  2.183519 seconds (2.97 M allocations: 221.639 MiB, 3.81% gc time, 27.94% compilation time)
```

### DIFF
--- a/src/VideoIO.jl
+++ b/src/VideoIO.jl
@@ -30,7 +30,7 @@ include("encoding.jl")
 include("testvideos.jl")
 using .TestVideos
 
-@static if Sys.islinux()
+if Sys.islinux()
     import Glob
     function init_camera_devices()
         append!(CAMERA_DEVICES, Glob.glob("video*", "/dev"))
--- a/src/VideoIO.jl
+++ b/src/VideoIO.jl
@@ -30,7 +30,7 @@ include("encoding.jl")
 include("testvideos.jl")
 using .TestVideos
 
-if Sys.islinux()
+@static if Sys.islinux()
     import Glob
     function init_camera_devices()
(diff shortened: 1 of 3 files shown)


src/VideoIO.jl
```diff
@@ -30,7 +30,7 @@ include("encoding.jl")
 include("testvideos.jl")
 using .TestVideos
 
-if Sys.islinux()
+@static if Sys.islinux()
```
IanButterworth, line 33: Not needed
```suggestion
if Sys.islinux()
```

src/avio.jl
```diff
@@ -1015,12 +1015,23 @@ function get_camera_devices(ffmpeg, idev, idev_name)
...
+
+    if isnothing(device) || isnothing(format) || isnothing(options) || isempty(CAMERA_DEVICES)
+        # do this only if needed
+        init_camera_devices()
+        init_camera_settings()
+    end
+
+    device = something(device, DEFAULT_CAMERA_DEVICE[])
+    format = something(format, DEFAULT_CAMERA_FORMAT[])
+    options = something(options, DEFAULT_CAMERA_OPTIONS)
```
IanButterworth, line 1033:
```suggestion
    device = @something(device, DEFAULT_CAMERA_DEVICE[])
    format = @something(format, DEFAULT_CAMERA_FORMAT[])
    options = @something(options, DEFAULT_CAMERA_OPTIONS)
```

PR #352, opened 2022-04-15
Fast encode path when image data layout matches ffmpeg buffer

IanButterworth:
When the source image is packed in the same way as the ffmpeg buffer, sending the data across can happen in one instruction, which may provide some speedup.


galenlynch (commented):
Looks great!

PR #368, opened 2022-06-06
Add tip on using hyphenated parameters

### LINKED ISSUE #365: Encoder options containing hyphens
For at least the libx265 codec, it uses a hyphenated argument, `x265-params`; this is required to pass certain options, such as `lossless=1`.

https://trac.ffmpeg.org/wiki/Encode/H.265

However, symbols containing hyphens cannot be used as named tuple names, which is used by `encoder_options`.

EDIT: `Symbol("x265-params")` isn't allowed, but `var"x265-params"` works. Not sure what to do about that here, but maybe the error message on the Julia side could be improved.

BioTurboNick:
Fixes #365


IanButterworth (approved):
Great. Thanks

Issue #371, opened 2022-07-07
`sigatomic_end` throws error with VSCode Debugger


mrufsvold:
`VideoIO.get_time_duration()` runs normally when I use it from the Julia REPL. However, when using the Julia Debugger in VSCode, it throws the following exception

```
sigatomic_end called in non-sigatomic region

Stacktrace:
 [1] close(avin::VideoIO.AVInput{String})
   @ VideoIO C:\Users\mrufsvold\.julia\packages\VideoIO\V24Bg\src\avio.jl:968
 [2] get_time_duration(file::String)
   @ VideoIO C:\Users\mrufsvold\.julia\packages\VideoIO\V24Bg\src\info.jl:45
```

It appears that PyCall had the [same problem](https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/219), and they were able to resolve it  by replacing[ `sigatomic_(end|begin)` with `disable_sigint`](https://github.com/JuliaPy/PyCall.jl/pull/686). My understanding from reading their conversation is that `sigatomic_(end|begin)` is an internal function for Julia and `disable_sigint` is its the public API sibling.

I can put in a PR if the maintainers aren't opposed to the change for some reason!

IanButterworth:
Closed by #372

PR #372, opened 2022-07-07
Replace sigatomic_(begin|end) with disable_sigint() do ... end

mrufsvold:
Based on [this issue](https://github.com/JuliaIO/VideoIO.jl/issues/371), I switched out all of the uses of sigatomic with disable_sigint. Because the functions are now wrapped in a do block, I had to define `ret` as a `Ref{Int32}` outside the block to make it accessible to the error catches. 

I tested `load()`, `get_number_frames()`, and `get_duration()` both in the REPL and using the VSCode debugger. They all worked in both cases now.

I'm not sure if there are other tests that should be run, but hopefully, someone more knowledgeable about this package and verify all is well!


src/avio.jl
```diff
@@ -116,10 +116,13 @@ convert(::Type{AVRational}, r::Rational) = AVRational(numerator(r), denominator(
...
-        ret = av_read_frame(avin.format_context, avin.packet)
-        sigatomic_end()
-        if ret < 0
+        # Need to declare ret outside the do block
+        ret = Ref{Int32}()
+        disable_sigint() do
+            ret[] = av_read_frame(avin.format_context, avin.packet)
+        end
+
+        if ret[] < 0
```
IanButterworth, line 125:
```suggestion
        ret = disable_sigint() do
            av_read_frame(avin.format_context, avin.packet)
        end
        if ret < 0
```

src/avio.jl
```diff
@@ -296,12 +299,13 @@ function VideoReader(
...
-    unlock(VIO_LOCK)
-    sigatomic_end()
-    ret < 0 && error("Could not open codec")
+    ret = Ref{Int32}()
+    disable_sigint() do
+        lock(VIO_LOCK)
+        ret[] = avcodec_open2(codec_context, codec, C_NULL)
+        unlock(VIO_LOCK)
+    end
+    ret[] < 0 && error("Could not open codec")
```
IanButterworth, line 308:
```suggestion
    ret = disable_sigint() do
        lock(VIO_LOCK) do
            avcodec_open2(codec_context, codec, C_NULL)
        end
    end
    ret < 0 && error("Could not open codec")
```

src/encoding.jl
```diff
@@ -324,12 +324,13 @@ options, or pass the private options to `encoder_private_options` explicitly""",
...
-    unlock(VIO_LOCK)
-    sigatomic_end()
-    ret < 0 && error("Could not open codec: Return code $(ret)")
+    ret = Ref{Int32}()
+    disable_sigint() do
+        lock(VIO_LOCK)
+        ret[] = avcodec_open2(codec_context, codec, C_NULL)
+        unlock(VIO_LOCK)
+    end
+    ret[] < 0 && error("Could not open codec: Return code $(ret)")
```
IanButterworth, line 333:
```suggestion
    ret = disable_sigint() do
        lock(VIO_LOCK) do
            avcodec_open2(codec_context, codec, C_NULL)
        end
    end
    ret < 0 && error("Could not open codec: Return code $(ret)")
```

src/frame_graph.jl
```diff
@@ -134,18 +134,19 @@ function exec!(s::SwsTransform)
...
+                s.sws_context,
+                src_data_ptr,
+                src_linesize_ptr,
+                0,
+                s.dstframe.height,
+                dst_data_ptr,
+                dst_linesize_ptr,
+            )
+        end
+        ret[] <= 0 && error("Could not scale frame")
```
IanButterworth, line 149:
```suggestion
        ret = disable_sigint() do
            sws_scale(
                s.sws_context,
                src_data_ptr,
                src_linesize_ptr,
                0,
                s.dstframe.height,
                dst_data_ptr,
                dst_linesize_ptr,
            )
        end
        ret <= 0 && error("Could not scale frame")
```

.gitignore
```diff
@@ -10,3 +10,4 @@ test/precisiontest_gray_test.mp4
 test/ordertest_gray_test.mp4
 docs/build/
 test/temp.stream
+.vscode/
```
IanButterworth, line 13:
```suggestion
.vscode/

```

IanButterworth (changes requested):
Thanks for this.

I don't think the `Ref`'s are needed? Also the locks can be tidied up

mrufsvold:
Addressed your feedback!

src/avio.jl
```diff
@@ -116,9 +116,11 @@ convert(::Type{AVRational}, r::Rational) = AVRational(numerator(r), denominator(
 # Pump input for data
 function pump(avin::AVInput)
     while avin.isopen && !avin.finished
-        sigatomic_begin()
-        ret = av_read_frame(avin.format_context, avin.packet)
-        sigatomic_end()
+        # Need to declare ret outside the do block
```
IanButterworth, line 119:
```suggestion
```

IanButterworth (approved):
Thanks. Sorry for the delay

IanButterworth:
There seems to be a problem with nightly. Adding 1.8-nightly to evaluate.

I'll merge either way and fix on another PR

PR #385, opened 2023-01-04
Remove Requires-based GLMakie video player functions

IanButterworth:
Compared to the Images ecosystem, which has ImageIO & ImageView / ImageInTerminal etc. it's never made too much sense that the Makie-based video player lives in here, even if it is behind Requires.

I regret not doing this before v1.0.0 because I think this means we'd need to release v2.0.0 after only a few patch versions.

Before releasing this, this functionality needs to be moved to another package.


timholy:
Removing or refactoring makes sense simply from the standpoint of precompilation (though an extension module would fix that).

If you're thinking about 2.0, have you also considered further changes? Video data seems array-like, with one or two caveats:
- in some applications (e.g., accessing a movie file), for `img[i, j, k]`, updating `k` is much more "expensive" than `i` or `j`
- in some applications (e.g., a live webcam feed), you can't control `k` independently

The first can be viewed as something that could be expressed with a trait and thus might not be a hard-break of an array abstraction; the second makes video less clearly array-like. Are there enough other applications that these concepts would be worth encapsulating more abstractly? I attempted to do this with [ImageAxes's `StreamingContainer`](https://github.com/JuliaImages/ImageAxes.jl/blob/2cde1bfcadf70403ae3b2eec93ad91c5fa3a8017/src/ImageAxes.jl#L191-L265), but I don't think it's been widely adopted. There may be a better platform available now?

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+VideoIO v1.1.0 Release Notes
+======================
+## Removal
+
+- The GLMakie-based video player that was accessed through Requires by loading GLMakie separately has been removed
+  after being deprecated in v1.0.7.
```
IanButterworth, line 6:
```suggestion
  after being deprecated in v1.0.8.
```

PR #393, opened 2023-02-08
Add VideoIO.framerate(f) to the manual and expand tests

IanButterworth:
Related to #392 

cc. @AbelHo

PR #394, opened 2023-02-11
fixed test result for frame rate (fps)

AbelHo:
I have edited the correct fps that the test should check for.

I am guessing you prefer to fix this via julia instead of using ffprobe?

The framerate now for
Ladybird opening wings CC-BY NatureClip.mp4 -> 30000//1
it should be
30000//1001

Artist’s_impression_of_the_black_hole_inside_NGC_300_X-1_(ESO_1004c).webm -> 1000//1
should be
25/11


IanButterworth:
I'll merge this into #393 and continue debugging over there

PR #409, opened 2023-11-16
JET fixes/findings

IanButterworth:
@galenlynch I've fixed the obvious things and left comments for the less obvious, if you can help.

To run this just
```
using JET
report_package("VideoIO")
```

PR #418, opened 2024-02-17
Update to FFMPEG_jll v6.1.1. Minor bump to v1.2.0

theogf:
As suggested in https://github.com/JuliaIO/FFMPEG.jl/pull/59#discussion_r1492764393, this should test if VideoIO runs correctly with the new version o FFMPEG


.github/workflows/CI.yml
```diff
@@ -53,6 +53,13 @@ jobs:
           version: ${{ matrix.julia-version }}
       - uses: julia-actions/cache@v1
       - uses: julia-actions/julia-runtest@v1
+      - uses: julia
+      - name: Configure new FFMPEG version environment
+        shell: julia --project="." --color=yes {0}
+        run: |
+          using Pkg
+          Pkg.add(PackageSpec(url="https://github.com/theogf/FFMPEG.jl/tree/tgf/update-FFMPEG"))
+          Pkg.instantiate()
```
giordano, line 62: I presume you want to install FFMPEG _before_ running the tests, no?
```suggestion
      - name: Configure new FFMPEG version environment
        shell: julia --project="." --color=yes {0}
        run: |
          using Pkg
          Pkg.add(PackageSpec(url="https://github.com/theogf/FFMPEG.jl/tree/tgf/update-FFMPEG"))
          Pkg.instantiate()
      - uses: julia-actions/julia-runtest@v1
```

theogf: Indeed! I thought I did sorry.

giordano:
For the benefit of future (and present) readers, the [segmentation fault](https://github.com/JuliaIO/VideoIO.jl/actions/runs/7941832646/job/21684548855?pr=418#step:6:307) @theogf refers to is
```
[6222] signal (11.2): Segmentation fault: 11
in expression starting at /Users/runner/work/VideoIO.jl/VideoIO.jl/src/VideoIO.jl:203
av_freep at /Users/runner/.julia/artifacts/112745bebf2bab1be3313a35fda52d857c5b4ab4/lib/libavutil.58.29.100.dylib (unknown line)
avcodec_parameters_from_context at /Users/runner/.julia/artifacts/112745bebf2bab1be3313a35fda52d857c5b4ab4/lib/libavcodec.60.31.102.dylib (unknown line)
```
In particular, it's happening to https://github.com/JuliaIO/VideoIO.jl/blob/0398aa6be81a1e637132a39136ebad94afd2f196/src/VideoIO.jl#L208  My loose understanding is that VideoIO use only CLI `ffmpeg`, so it shouldn't be too complicated to craft a reproducible example to report upstream for someone motivated to push for FFMPEG v6, otherwise we'd have to yank that version, if it's broken (and will also be stuck forever to v4 until someone else does that anyway)

giordano:
Actually, I was looking for `ccall` in `src/`, but I just realised that there are thousands of `ccall`s in https://github.com/JuliaIO/VideoIO.jl/blob/0398aa6be81a1e637132a39136ebad94afd2f196/lib/libffmpeg.jl, I guess you'd need to update the generated wrappers, too.  Side note, this means that `VideoIO` pretty much needs to specify `FFMPEG_jll` explicitly.

giordano:
@theogf bump

theogf:
> Actually, I was looking for `ccall` in `src/`, but I just realised that there are thousands of `ccall`s in https://github.com/JuliaIO/VideoIO.jl/blob/0398aa6be81a1e637132a39136ebad94afd2f196/lib/libffmpeg.jl, I guess you'd need to update the generated wrappers, too. Side note, this means that `VideoIO` pretty much needs to specify `FFMPEG_jll` explicitly.

I am not sure what you mean here.
One wrapper is for example
```
function avcodec_decode_subtitle2(avctx, sub, got_sub_ptr, avpkt)
    ccall((:avcodec_decode_subtitle2, libavcodec), Cint, (Ptr{AVCodecContext}, Ptr{AVSubtitle}, Ptr{Cint}, Ptr{AVPacket}), avctx, sub, got_sub_ptr, avpkt)
end
```

Is the idea to just make sure that `libavcodec` directly comes from `FFMPEG_jll`? Or by update you mean to make sure that all wrappers are compatible with 6.1?

giordano:
> Or by update you mean to make sure that all wrappers are compatible with 6.1?

Yes, but you don't do that manually, but run the script in https://github.com/JuliaIO/VideoIO.jl/tree/42b25a21c0ec524d75bfead8209a0bfd61ca91b5/gen

theogf:
So I generated `libffmpeg.jl` again, (also restricting the dependency to FFMPEG_jll). However I get the error 
```
Failed to precompile VideoIO [d6d074c3-1acf-5d4c-9a43-ef38773959a2] to "/home/theo/.julia/compiled/v1.10/VideoIO/jl_nKnqkE".
ERROR: LoadError: UndefVarError: `AV_CHANNEL_LAYOUT_MASK` not defined
```

I located the corresponding C code here 
https://github.com/FFmpeg/FFmpeg/blob/84e669167de9394e0de1bee0244f48ffac6f3826/libavutil/channel_layout.h#L378

But I don't know how to translate this to Julia? (and add it to `prologue.jl`)

giordano:
That's a macro, it doesn't really have a correspondence in julia since it's a source-code transformation, not part of the ABI of the library.  I don't know what's going on there.

theogf:
> That's a macro, it doesn't really have a correspondence in julia since it's a source-code transformation, not part of the ABI of the library. I don't know what's going on there.

Ah I think it's just some constructor for `AVChannelLayout` which is defined just above.

giordano:
@Gnimuc do you have a clue how to deal with this?  `Clang.jl` correctly defined the macros
https://github.com/JuliaIO/VideoIO.jl/blob/a9384d99cf0b659339f3380516b62a530ae18bd2/lib/libffmpeg.jl#L22098-L22104 etc., but the macro [`AV_CHANNEL_LAYOUT_MASK`](https://github.com/FFmpeg/FFmpeg/blob/84e669167de9394e0de1bee0244f48ffac6f3826/libavutil/channel_layout.h#L378-L382) has not been expanded.

Gnimuc:
```
AV_CHANNEL_LAYOUT_MASK(nb, m) \
    { /* .order */ AV_CHANNEL_ORDER_NATIVE, \
      /* .nb_channels */  (nb), \
      /* .u.mask */ { m }, \
      /* .opaque */ NULL }
```

unfortunately, Clang.jl doesn't support such "convoluted" macros. Honestly, I don't even know how to manually translate this function-like macro into Julia... 

if there is no direct use, it's safe to ignore all of the `AV_CHANNEL*` macro identifiers via the .toml file.

giordano:
> if there is no direct use, it's safe to ignore all of the `AV_CHANNEL*` macro identifiers via the .toml file.

Good point, it looks like all the `AV_CHANNEL_LAYOUT_*` macros are only defined in the resulting file, but not used.  Thanks!

theogf:
Just a small update, adding the right regex worked, but now there are some deprecated functions I need find replacement for, namely `av_stream_get_r_frame_rate`

theogf:
Hopefully it should work now :sweat_smile:

Project.toml
```diff
@@ -20,8 +20,8 @@ Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 [compat]
 ColorTypes = "0.9, 0.10, 0.11"
 Downloads = "1.3"
-FFMPEG = "0.3, 0.4"
-FFMPEG_jll = "4.1"
+FFMPEG = "0.3, 0.4, 1"
+FFMPEG_jll = "4.1, 6.1"
```
IanButterworth, line 24:
```suggestion
FFMPEG_jll = "6.1"
```

gen/Project.toml
```diff
@@ -1,8 +1,8 @@
 [deps]
 Clang = "40e3b903-d033-50b4-a0cc-940c62c95e31"
-FFMPEG = "c87230d0-a227-11e9-1b43-d7ebe4e7570a"
+FFMPEG_jll = "b22a6f82-2f65-5046-a5b2-351ab43fb4e5"
 Match = "7eb4fadd-790c-5f42-8a69-bfa0b872bfbf"
 Vulkan_Headers_jll = "8d446b21-f3ad-5576-a034-752265b9b6f9"
 
 [compat]
-Clang = "0.14.0"
+Clang = "0.14, 0.15, 0.16, 0.17"
```
giordano, line 8: Also compat for ffmpeg_jll

src/avio.jl
```diff
@@ -832,10 +832,10 @@ get_frame_presentation_time(stream, frame) = convert(Rational, stream.time_base)
 function get_frame_period_timebase(r::VideoReader)
     # This is probably not valid for many codecs, frame period in timebase
     # units
-    stream = get_stream(s)
+    stream = get_stream(r)
```
giordano, line 835: This function was untested?

theogf: Looks like it :sweat_smile:

src/avio.jl
```diff
@@ -858,7 +858,7 @@ function seek_trim(r::VideoReader, seconds::Number)
     time_base = convert(Rational, stream.time_base)
     time_base == 0 && error("No time base")
     target_pts = seconds_to_timestamp(seconds, time_base)
-    frame_rate = convert(Rational, av_stream_get_r_frame_rate(stream))
+    frame_rate = convert(Rational, stream.r_frame_rate)
```
theogf, line 861: Tests are failing so I can only assume that this is not the solution?

src/avio.jl
IanButterworth: Yeah there's a few things like this in https://github.com/JuliaIO/VideoIO.jl/pull/409

src/avio.jl
IanButterworth: This seems like the correct change based on https://patchwork.ffmpeg.org/project/ffmpeg/patch/20171023160821.6384-2-jamrial@gmail.com/

And the test failures could be rounding assumptions. Not looked deeper into it
```
Encoding with frame rate 29.5: Test Failed at /home/runner/work/VideoIO.jl/VideoIO.jl/test/writing.jl:196
  Expression: parse(Float64, measured_dur_str[1]) == target_dur
   Evaluated: 3.389831 == 3.39
```
```
Encoding with frame rate 29.5: Test Failed at /home/runner/work/VideoIO.jl/VideoIO.jl/test/writing.jl:214
  Expression: parse(Float64, measured_dur_str[1]) == target_dur
   Evaluated: 3.389831 == 3.39
```

theogf: We're a bit far from rounding error? Maybe it's a difference of 1 frame?

IanButterworth:
It'd be great to get this figured out

src/avio.jl
IanButterworth: I think this is just correct
```
julia> 100 / 29.5
3.389830508474576

julia> 99 / 29.5
3.3559322033898304

julia> 101 / 29.5
3.4237288135593222
```

Perhaps `ffmpeg -v error -show_entries format=duration -of default=noprint_wrappers=1:nokey=1` just returns more precision on newer versions.

IanButterworth: I've pushed a change to the full precision

Project.toml
```diff
@@ -17,8 +17,8 @@ Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 [compat]
 ColorTypes = "0.9, 0.10, 0.11, 0.12"
 Downloads = "1.3"
-FFMPEG = "0.3, 0.4"
-FFMPEG_jll = "4.1"
+FFMPEG = "0.3, 0.4, 1"
+FFMPEG_jll = "6.1"
```
IanButterworth, line 21: Do we still need to directly depend on both?

giordano: https://github.com/JuliaIO/VideoIO.jl/pull/418#issuecomment-1950211326

Project.toml
```diff
@@ -17,8 +17,8 @@ Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 [compat]
 ColorTypes = "0.9, 0.10, 0.11, 0.12"
 Downloads = "1.3"
-FFMPEG = "0.3, 0.4"
-FFMPEG_jll = "4.1"
+FFMPEG = "0.3, 0.4, 1"
```
IanButterworth, line 20:
```suggestion
FFMPEG = "0.4.3"
```
https://github.com/JuliaIO/FFMPEG.jl/pull/59/files#r1896927588

test/writing.jl
```diff
@@ -193,14 +193,14 @@ end
...
-        @test parse(Float64, measured_dur_str[1]) == target_dur
+        @test parse(Float64, measured_dur_str[1]) ≈ target_dur
     end
 end
 
 @testset "Encoding video with float frame rates" begin
     n = 100
     fr = 29.5 # 59 // 2
-    target_dur = 3.39
+    target_dur = 3.389831
```
giordano, line 203: Should this be n / fr, or there's a good reason to spell the number out explicitly?

giordano (commented):
> I think we can move forward, do you agree?

Yes.

test/writing.jl
IanButterworth:
```suggestion
    target_dur = n / fr
```

IanButterworth:
FFMPEG.jl v0.4.3 registered https://github.com/JuliaRegistries/General/pull/135435

PR #452, opened 2026-04-29
fix compiling VideoIO with JuliaC

bjarthur:
here is a minimal working example for which compilation fails:

```
module FFMPEGTest

using VideoIO

(@main)(args) = return 0

end # module FFMPEGTest
```

`]generate FFMPEGTest` and put the above in src/FFMPEGTest.jl.

then `]app add JuliaC@0.2.3`

finally `~/.julia/bin/juliac --output-exe foo --bundle build FFMPEGTest`

all sort of errors result.  this PR fixes it.


lib/libffmpeg.jl
```diff
@@ -14015,7 +14015,31 @@ struct AVAMFDeviceContext
...
+AMF_SURFACE_R10G10B10A2 = 13
+AMF_SURFACE_Y210 = 14
+AMF_SURFACE_AYUV = 15
+AMF_SURFACE_Y410 = 16
+AMF_SURFACE_Y416 = 17
+AMF_SURFACE_GRAY9 = 18
+AMF_SURFACE_GRAY10 = 19
+AMF_SURFACE_GRAY12 = 20
+AMF_SURFACE_GRAY14 = 21
+AMF_SURFACE_GRAY16 = 22
```
giordano: Feels like these should be enums, not (non-`const`) globals

giordano: To be clear, enum is a thing in Julia: https://docs.julialang.org/en/v1/base/base/#Base.Enums.@enum

bjarthur: oh!  yes.  just changed it to conform to how clang handles enums elsewhere in this file.  e.g. https://github.com/JuliaIO/VideoIO.jl/blob/master/lib/libffmpeg.jl#L143-L155

PR #453, opened 2026-05-26
Add `strict` mode for video corruption detection

IanButterworth:
Add a `strict::Bool = false` keyword to `openvideo` / `VideoReader`. When enabled, the underlying codec context is opened with `AV_EF_EXPLODE | AV_EF_CRCCHECK | AV_EF_BITSTREAM`, so bitstream / CRC errors propagate out of the decoder instead of being silently concealed. These errors surface to the caller as a `VideoCorruptionError`.

This makes decoding more deterministic and catches corruption for callers that need data integrity (no synthetic pixels from error concealment), while the default behaviour is unchanged.

## Implementation

- Export `VideoCorruptionError`
- Set `err_recognition` flags before `avcodec_open2` when `strict=true`
- Check both `avcodec_send_packet` and `avcodec_receive_frame` return codes for `AVERROR_INVALIDDATA`
- After successful frame decode, check `decode_error_flags` and `AV_FRAME_FLAG_CORRUPT` to catch frames marked corrupt by the decoder
- Throw `VideoCorruptionError` for all detected corruption in strict mode
- Documentation in `docs/src/reading.md` and `CHANGELOG.md`
- Tests covering the new keyword, exception type, normal-mode pass-through, and corruption detection via a malformed H.264 Annex-B bitstream

## Robustness

The implementation checks multiple corruption signals as recommended by FFmpeg maintainers:
- Return codes from both packet sending and frame receiving
- `decode_error_flags` on decoded frames
- `AV_FRAME_FLAG_CORRUPT` flag on decoded frames

Note: Corruption detection remains codec/container dependent; FFmpeg's error recognition is best-effort.

### DIFF
--- a/src/avio.jl
+++ b/src/avio.jl
@@ -735,8 +735,10 @@ function decode(r::VideoReader, packet)
     elseif fret == VIO_AVERROR_EOF
         r.finished = true
     elseif fret != -Libc.EAGAIN
-        r.strict && throw(VideoCorruptionError(
-            "Decoder reported a bitstream error: $(av_error_string(fret))"))
+        # In strict mode, throw VideoCorruptionError specifically for bitstream/corruption errors
+        if r.strict && fret == libffmpeg.AVERROR_INVALIDDATA
+            throw(VideoCorruptionError("Bitstream corruption detected: $(av_error_string(fret))"))
+        end
         error("Decoding error: $(av_error_string(fret))")
     end
     if !r.finished && !r.flush && pret == -Libc.EAGAIN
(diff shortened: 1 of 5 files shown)


src/avio.jl
```diff
@@ -711,6 +735,8 @@ function decode(r::VideoReader, packet)
     elseif fret == VIO_AVERROR_EOF
         r.finished = true
     elseif fret != -Libc.EAGAIN
+        r.strict && throw(VideoCorruptionError(
```
Octogonapus: I am not sure this is correct. If the `error` below ever ran before this patch, then there are some errors that could happen with `strict=true` that are not bitstream errors.

Issue #427, opened 2025-01-02
Inaccurate seek


yakir12:
I have a video where `seek` yields highly inaccurate results. Maybe this can be explained by some "quality" of the video, but if so, it might be worth mentioning in `seek`'s docstring what the accuracy of `seek` depends on.

MWE (note that the link here is valid only for the next 7 days, expiring approximately on 2025-01-09 23:00:00 CET): 

```julia
using VideoIO

url = "https://vision-group-temporary.s3.eu-central-1.amazonaws.com/00002_K02.MTS"
file = download(url) # sorry, 740 MB, takes a moment
vid = openvideo(file)
t = 2 # arbitrary time stamp, feel free to choose another
seek(vid, t)
read(vid)
tt = gettime(vid) # 2.46 >> 2
close(vid)

# the following proves there are many frames closer to the desired target time point:
vid = openvideo(file)
ts = Float64[gettime(vid)]
while ts[end] < tt # collect up to the claimed time stamp
    read(vid)
    push!(ts, gettime(vid))
end
close(vid)

filter(≥(t), ts) # 24 frames between the desired and claimed time points
```

I'm on `VideoIO v1.1.1` and Julia Version 1.11.2.

yakir12:
Here, I calculate the intended time stamp and the actual one for every second in the first minute of the video:
```julia
t0 = 1:60
t1 = map(t0) do t
    openvideo(file) do vid
        seek(vid, t)
        gettime(vid)
    end
end
```
These are the differences:
```julia
julia> lineplot(t0 .- t1)
        ┌────────────────────────────────────────┐ 
      0 │⢸⠀⠀⠀⠀⠀⠀⠀⣾⠀⠀⠀⠀⠀⠀⠀⣾⠀⠀⠀⠀⠀⠀⠀⣾⠀⠀⠀⠀⠀⠀⠀⣾⠀⠀⠀⠀⠀⠀⠀│ 
        │⢸⠀⠀⠀⠀⠀⠀⢰⢹⠀⠀⠀⠀⠀⠀⢰⢹⠀⠀⠀⠀⠀⠀⢰⢹⠀⠀⠀⠀⠀⠀⢰⢹⠀⠀⠀⠀⠀⠀⢰│ 
[...]

PR #469, opened 2026-09-16
Trim a seek by the measured frame period, not r_frame_rate (#468)

### LINKED ISSUE #468: seek lands one frame late on field-coded interlaced video (r_frame_rate is the field rate)
Follow-up to #427 / #467. On **field-coded interlaced** H.264 (PAFF, which AVCHD/HDV camcorders write), `seek` lands one frame late whenever the target falls between two frames.

`seek_trim` takes the frame period from `stream.r_frame_rate`. On these streams that is the **field** rate, twice the rate at which decoded frames arrive. `frame.duration` reports the field period too. Here's a 50i camcorder `.MTS` file:

```
ffprobe: field_order=tt  r_frame_rate=50/1  avg_frame_rate=25/1
time_base 1/90000:  r_frame_rate period = 1800,  frame.duration = 1800,  decoded pts gaps = 3600
```

`seek_trim` drops a frame while `frame.pts + period <= target`. With a period of half a frame, it also drops the frame that actually holds the target whenever the target is in the second half of that frame's interval. On VideoIO `master` (after #467), `seek(v, t0 + s); read!(v, img); gettime(v)` compared with a sequential decode:

| s | correct frame | `seek` gives |
|---|---|---|
| 0.02 | t0 + 0.00 | t0 + 0.04 |
| 0.5 | t0 + 0.48 | t0 + 0.52 |
| 1.3 | t0 + 1.28 | t0 + 1.32 |
| 7.3 | t0 + 7.28 | t0 + 7.32 |

Targets that fall exactly on a frame are correct, and progressive files are unaffected: a 50p `.MTS`, MPEG-TS/MP4 from libx264, and camera MP4s all check out.

I couldn't make a synthetic reproduction with ffmpeg. libx264 and mpeg2video interlaced encodes, VUI tick-rate rewrites, remuxes and timestamp shifts all come out with `r_frame_rate` equal to the real frame rate. I can check candidate files against the real footage. A PR follows.

yakir12:
Fixes #468.

## Problem

`seek_trim` takes the frame period from `stream.r_frame_rate`. On field-coded interlaced H.264 (PAFF, from AVCHD/HDV camcorders), that is the *field* period, half the real frame period, and `frame.duration` is no better. So a target in the second half of a frame's interval dropped the frame that holds it, and the seek landed one frame late. Details and measurements are in #468.

## Fix

The video stream is no longer trimmed with `r_frame_rate`. After the back-off from #467 lands at or before the target, `seek` walks forward using the **pts gap between consecutive decoded frames** as the period:
- It keeps a frame once the next frame's interval would start past the target.
- A frame is only known to be the right one after the following frame has been decoded. When that following frame is already past the target, `seek` seeks to the same point again, which deterministically lands on the same frame, and stops there. If it doesn't land on the same frame, it falls back to `seek_trim`.

The seek-and-decode-first-frame step is factored into `seek_landing_pts!`, now used by both the back-off loop and the re-seek. Other streams in the container are still trimmed with `seek_trim`, as before.

Cost: at most one extra seek, which only happens when the frame the seek lands on is itself the target frame.

## Tests

- Full suite `Pkg.test()` passes: 1562/1562 on Julia 1.13, including the #427 MPEG-TS seek test.
- Checked against a sequential decode, with 15 targets per file on and between frames, plus targets before the stream and `seekstart`. All land on the frame whose `[pts, pts + period)` holds the target:
  - a PAFF 50i `.MTS`: previously one frame late on 5 of 15 targets
  - a 50p `.MTS`
  - two 25 fps AVCHD `.MTS` files
  - synthetic libx264 MPEG-TS and MP4
  - a camera MP4

There is **no new regression test**, because I couldn't generate a stream whose `r_frame_rate` is the field rate. libx264 and mpeg2video interlaced encodes, VUI tick-rate rewrites, remuxing and timestamp shifts all came out with `r_frame_rate` equal to the real frame rate. If you know a recipe, or want a short PAFF clip added to the test videos, I'm happy to add the test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


IanButterworth:
Fable figured out a test so I pushed it

IanButterworth (approved):
Windows failure is unrelated windows FS stuff

